PR #35, opened 2022-12-27
Abstract ElasticClient (auto-provide executor with .run)

arnoldlacko:


arnoldlacko:
Closing, as this will be solved by adding a facade layer.

arnoldlacko:
Reopening, as the discussion is still relevant. :)

modules/library/src/main/scala/zio/elasticsearch/ElasticClient.scala
```diff
@@ -0,0 +1,11 @@
+package zio.elasticsearch
+
+import zio.{Task, ZLayer}
+
+abstract class ElasticClient(executor: ElasticExecutor) {
+
+  protected implicit class RunnableRequest[A](request: ElasticRequest[A]) {
+    def run: Task[A] = request.execute.provideLayer(ZLayer.succeed(executor))
```
mijicd, line 8: The downside of this approach is constant `provide`.

arnoldlacko: I was under the impression that having to `provide` is only a developer inconvenience, I missed the fact that there is a performance cost. I'll close this now.

PR #92, opened 2023-02-09
Use module pattern 2.0 (move requests to executor)

mvelimir:


modules/example/src/main/scala/example/Main.scala
```diff
@@ -46,14 +46,14 @@ object Main extends ZIOAppDefault {
     val deleteIndex: RIO[ElasticExecutor, Unit] =
       for {
         _ <- ZIO.logInfo(s"Deleting index '$Index'...")
-        _ <- ElasticRequest.deleteIndex(Index).execute
+        _ <- ZIO.serviceWithZIO[ElasticExecutor](_.deleteIndex(Index).execute)
```
mijicd, line 49: It would be great not having needing the call to `execute` as it makes the API usage awkward.

mvelimir: The only idea that comes to mind is an implicit conversion from `ElasticRequest` to `ZIO`.

PR #95, opened 2023-02-20
(dsl): Support Aggregations

dbulaja98:


modules/example/src/main/scala/example/RepositoriesElasticsearch.scala
```diff
@@ -32,7 +32,9 @@ import zio.prelude.Newtype.unsafeWrap
 final case class RepositoriesElasticsearch(executor: ElasticExecutor) {
 
   def findAll(): Task[List[GitHubRepo]] =
-    executor.execute(ElasticRequest.search[GitHubRepo](Index, matchAll))
+    executor
+      .execute(ElasticRequest.search[GitHubRepo](index = Index, query = matchAll, maybeAggregation = None))
```
mvelimir, line 36: Could this be renamed to just `aggregation`, and have `None` as the default parameter? Then we would use a method on `Search` requests to specify it.

modules/library/src/main/scala/zio/elasticsearch/AggregationResponse.scala
```diff
@@ -0,0 +1,46 @@
...
+object BucketAggregationResponse {
+  implicit val schema: Schema[BucketAggregationResponse] = DeriveSchema.gen[BucketAggregationResponse]
+}
+
+sealed trait AggregationBucket
+
+final case class TermsAggregationBucket(
+  key: String,
+  @fieldName("doc_count")
+  docCount: Int
```
drmarjanovic, line 41: Let's use `count` here.

modules/library/src/main/scala/zio/elasticsearch/AggregationResponse.scala
```diff
@@ -0,0 +1,46 @@
...
+  @fieldName("sum_other_doc_count")
+  sumOtherDocCount: Int,
+  buckets: List[TermsAggregationBucket]
+) extends AggregationResponse
+
+object BucketAggregationResponse {
+  implicit val schema: Schema[BucketAggregationResponse] = DeriveSchema.gen[BucketAggregationResponse]
+}
+
+sealed trait AggregationBucket
```
drmarjanovic, line 36: Do we need this?

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregation.scala
```diff
@@ -0,0 +1,58 @@
...
+      case m: MultipleAggregations => ("aggs", m.paramsToJson) // nested aggregations
+    }: _*)
+  }
+
+  private[elasticsearch] object MultipleAggregations {
+
+    def empty: MultipleAggregations = MultipleAggregations(Nil)
+  }
+
+  private[elasticsearch] final case class Terms(name: String, field: String) extends ElasticAggregation {
```
drmarjanovic, line 54: Let's use `TermsAggregation`. We use that for queries as well.

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregation.scala
```diff
@@ -0,0 +1,58 @@
...
+      self.copy(aggregations = self.aggregations ++ aggregations)
+
+    def paramsToJson: Json = Obj(aggregations.map {
+      case Terms(name, field)      => (name, Obj("terms" -> Obj("field" -> field.toJson)))
+      case m: MultipleAggregations => ("aggs", m.paramsToJson) // nested aggregations
+    }: _*)
+  }
+
+  private[elasticsearch] object MultipleAggregations {
+
```
drmarjanovic, line 50: Let's remove this line.

modules/library/src/main/scala/zio/elasticsearch/ElasticPrimitive.scala
```diff
@@ -0,0 +1,47 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package zio.elasticsearch
+
+import zio.json.ast.Json
+import zio.json.ast.Json.{Num, Str}
+
+object ElasticPrimitive {
```
drmarjanovic, line 22: Can you extract the `sealed trait` above, and provide `ElasticPrimitive` with implicit under it?

modules/library/src/main/scala/zio/elasticsearch/ElasticRequest.scala
```diff
@@ -85,17 +87,63 @@ object ElasticRequest {
         Right(None)
     }
 
-  def search[A](index: IndexName, query: ElasticQuery[_])(implicit
+  def search(
+    index: IndexName,
+    aggregation: ElasticAggregation
+  ): ElasticRequest[ScalaMap[String, AggregationResponse], Search] =
```
drmarjanovic, line 93: Does it make sense to introduce `ElasticAggregations` or `Aggregations` that will hide the `Map` implementation?

modules/library/src/main/scala/zio/elasticsearch/ElasticRequest.scala
```diff
@@ -85,17 +87,63 @@ object ElasticRequest {
...
+    SearchRequest(index = index, query = None, aggregation = Some(aggregation), routing = None).map { response =>
+      Obj.decoder.decodeJson(response.aggregations.get.toString) match {
+        case Left(e) =>
```
drmarjanovic, line 96: I prefer `Right(...)` then `Left(...)` sorting.

modules/library/src/main/scala/zio/elasticsearch/ElasticRequest.scala
```diff
@@ -205,11 +253,17 @@ object ElasticRequest {
...
+  private[elasticsearch] final case class SearchRequest(
     index: IndexName,
-    query: ElasticQuery[_],
+    query: Option[ElasticQuery[_]],
+    aggregation: Option[ElasticAggregation],
     routing: Option[Routing]
-  ) extends ElasticRequest[ElasticQueryResponse, GetByQuery]
+  ) extends ElasticRequest[ElasticQueryResponse, Search]
+
+  private[elasticsearch] final case class SearchResponse[A](
```
drmarjanovic, line 263: Let's add the methods we were talking about... `getAggregation` or `aggregation(name: String)`/ Also, does it make sense to rename it to `ElasticResponse`?

drmarjanovic: I guess we should add `items` or `data` as well.

drmarjanovic (changes requested):

modules/library/src/main/scala/zio/elasticsearch/ElasticRequest.scala
```diff
@@ -205,11 +253,17 @@ object ElasticRequest {
     routing: Option[Routing]
   ) extends ElasticRequest[Option[Document], GetById]
 
-  private[elasticsearch] final case class GetByQueryRequest(
+  private[elasticsearch] final case class SearchRequest(
     index: IndexName,
-    query: ElasticQuery[_],
+    query: Option[ElasticQuery[_]],
```
arnoldlacko, line 258: We might want to think about separating into 3 case classes to represent only valid combinations, something like:
- SearchRequest
- AggregationRequest
- SearchWithAggregationRequest

arnoldlacko: Possibly having then the API:
- search(query)
- aggregate(aggregation)
- search(query).aggregate(aggregation)

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregation.scala
```diff
@@ -0,0 +1,58 @@
...
+  def terms(name: String, field: String): ElasticAggregation = Terms(name, field)
+
+  private[elasticsearch] final case class MultipleAggregations(aggregations: List[ElasticAggregation])
+      extends ElasticAggregation { self =>
+
+    def aggregations(aggregations: ElasticAggregation*): MultipleAggregations =
+      self.copy(aggregations = self.aggregations ++ aggregations)
+
+    def paramsToJson: Json = Obj(aggregations.map {
+      case Terms(name, field)      => (name, Obj("terms" -> Obj("field" -> field.toJson)))
```
arnoldlacko, line 44: Should this rely on `terms.paramsToJson` to avoid duplication?

modules/library/src/main/scala/zio/elasticsearch/AggregationResponse.scala
```diff
@@ -0,0 +1,46 @@
...
+
+final case class TermsAggregationBucket(
+  key: String,
+  @fieldName("doc_count")
+  docCount: Int
+) extends AggregationBucket
```
arnoldlacko, line 42: We might need to leave space for the results of sub-aggregations here 🤔

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregation.scala
```diff
@@ -0,0 +1,58 @@
...
+
+  def paramsToJson: Json
+
+  final def toJson: Json = Obj("aggs" -> paramsToJson)
+
+}
+
+object ElasticAggregation {
+
+  def multipleAggregations: MultipleAggregations = MultipleAggregations.empty
```
arnoldlacko, line 33: I wonder if instead of having to use `multipleAggregation.aggregations(...)` we should have something like `terms(...).with(otherAgg).sub(subAgg.with(otherSubAgg))`

modules/library/src/main/scala/zio/elasticsearch/ElasticRequest.scala
dbulaja98: We already have `ElasticGetResponse`, etc.. So i guess it can be renamed to `ElasticSearchResponse`?

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregation.scala
dbulaja98: No, because we have in `terms.paramsToJson` this: `Obj(name -> Obj("terms" -> Obj("field" -> field.toJson)))`, but here we need only tuple `(name, Obj("terms" -> Obj("field" -> field.toJson)))`.

modules/library/src/it/scala/zio/elasticsearch/HttpExecutorSpec.scala
```diff
@@ -272,13 +274,103 @@ object HttpExecutorSpec extends IntegrationSpec {
...
                   res <- ElasticExecutor.execute(ElasticRequest.search[CustomerDocument](firstSearchIndex, query))
-                } yield assert(res)(Assertion.contains(firstCustomer))
+                } yield assert(res.data)(Assertion.contains(firstCustomer))
             }
           } @@ around(
             ElasticExecutor.execute(ElasticRequest.createIndex(firstSearchIndex, None)),
             ElasticExecutor.execute(ElasticRequest.deleteIndex(firstSearchIndex)).orDie
           )
         ) @@ shrinks(0),
+        suite("search with Aggregations")(
```
drmarjanovic, line 284: I would use `search with aggregations`.

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregation.scala
```diff
@@ -0,0 +1,105 @@
...
+  def paramsToJson: Json
+
+  final def sub(aggregation: ElasticAggregation[_ <: SingleElasticAggregationType])(implicit
+    wsa: WithSubAggregation[EAT]
+  ): ElasticAggregation[EAT] =
+    wsa.sub(aggregation = self, subAggregation = aggregation)
+
+  final def toJson: Json = Obj("aggs" -> paramsToJson)
+
+  final def `with`(aggregation: ElasticAggregation[_ <: SingleElasticAggregationType])(implicit
```
drmarjanovic, line 37: @arnoldlacko should we use `withAggregation`?

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregation.scala
```diff
@@ -0,0 +1,105 @@
...
+import zio.elasticsearch.OtherAggregation.WithOtherAggregation
+import zio.elasticsearch.SubAggregation.WithSubAggregation
+import zio.json.ast.Json
+import zio.json.ast.Json.Obj
+
+sealed trait ElasticAggregation[EAT <: ElasticAggregationType] { self =>
+
+  def paramsToJson: Json
+
+  final def sub(aggregation: ElasticAggregation[_ <: SingleElasticAggregationType])(implicit
```
drmarjanovic, line 30: What about `withSubaggregation`?

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregation.scala
```diff
@@ -0,0 +1,105 @@
...
+  ): ElasticAggregation[Multiple] =
+    woa.`with`(aggregation = self, otherAggregation = aggregation)
+
+}
+
+object ElasticAggregation {
+
+  def multipleAggregations: MultipleAggregations = MultipleAggregations.empty
+
+  def termsAggregation(name: String, field: String): ElasticAggregation[Terms] = TermsAggregation(name, field)
```
drmarjanovic, line 48: @arnoldlacko, does it make sense to support `Class.field` syntax as well instead of `field: String`?

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregation.scala
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  private[elasticsearch] object MultipleAggregations {
+    def empty: MultipleAggregations = MultipleAggregations(Nil)
+  }
+
+  private[elasticsearch] final case class TermsAggregation(
+    name: String,
+    field: String,
+    subAggregations: List[ElasticAggregation[_ <: SingleElasticAggregationType]] = Nil
```
drmarjanovic, line 77: Do we use default arguments here?

modules/library/src/main/scala/zio/elasticsearch/TestExecutor.scala
```diff
@@ -28,6 +28,8 @@ private[elasticsearch] final case class TestExecutor private (data: TMap[IndexNa
 
   def execute[A](request: ElasticRequest[A, _]): Task[A] =
     request match {
+      case AggregationRequest(index, _, _) =>
```
drmarjanovic, line 31: This can be removed completely.

modules/library/src/test/scala/zio/elasticsearch/AggregationSpec.scala
```diff
@@ -0,0 +1,179 @@
...
+  multipleAggregations,
+  termsAggregation
+}
+import zio.elasticsearch.utils.RichString
+import zio.test.Assertion.equalTo
+import zio.test._
+
+object AggregationSpec extends ZIOSpecDefault {
+
+  override def spec: Spec[Environment with TestEnvironment with Scope, Any] =
```
drmarjanovic, line 16: Not sure do we use `override` in the rest of the tests.

drmarjanovic (changes requested):

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregation.scala
arnoldlacko: Yes, or I wonder if `withAgg` might be better to not make the querie longer than necessary  🤔

arnoldlacko: I far as I remember Elastic also uses `aggs` so it should be clear in enough what it means

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregation.scala
drmarjanovic: Or `withSubAggregation` or `withSubAgg`?

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregation.scala
arnoldlacko: I'm fine with either

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregation.scala
arnoldlacko: Sounds reasonable to me, perhaps @mvelimir can confirm.

modules/library/src/main/scala/zio/elasticsearch/AggregationResponse.scala
```diff
@@ -0,0 +1,87 @@
...
+          }
+        }.toMap
+
+        val key      = dynamicFields.get("key").asInstanceOf[Option[String]].get
+        val docCount = dynamicFields.get("doc_count").asInstanceOf[Option[Int]].get
+        val subAggs = dynamicFields.filter { case (field, _) =>
+          !field.equals("key") && !field.equals("doc_count")
+        }.values.toList.map(value => value.asInstanceOf[BucketAggregationResponse])
+
+        TermsAggregationBucket.apply(key, docCount, if (subAggs.isEmpty) None else Some(subAggs))
```
drmarjanovic, line 82: Maybe replace with `Option(subAggs).filter(_.nonEmpty)`?

modules/library/src/main/scala/zio/elasticsearch/AggregationResponse.scala
```diff
@@ -0,0 +1,87 @@
...
+                    .get
+                    .map(bucket => TermsAggregationBucket.decoder.decodeJson(bucket.toJson).toOption)
+                    .map(_.get)
+                )
+              )
+          }
+        }.toMap
+
+        val key      = dynamicFields.get("key").asInstanceOf[Option[String]].get
+        val docCount = dynamicFields.get("doc_count").asInstanceOf[Option[Int]].get
```
drmarjanovic, line 77: Why do we have `asInstanceOf[Option[Int]]` and then `.get`? 

Can we replace this with `dynamicFields("doc_count").asInstaceOf[Int]`?

modules/library/src/main/scala/zio/elasticsearch/AggregationResponse.scala
```diff
@@ -0,0 +1,87 @@
...
+        val subAggs = dynamicFields.filter { case (field, _) =>
+          !field.equals("key") && !field.equals("doc_count")
```
drmarjanovic, line 79: Why not use `"key" != field` and `"doc_count" != field`?

modules/library/src/main/scala/zio/elasticsearch/AggregationResponse.scala
```diff
@@ -0,0 +1,87 @@
...
+        }.values.toList.map(value => value.asInstanceOf[BucketAggregationResponse])
```
drmarjanovic, line 80: Can we use `_.asInstanceOf[...]`?

modules/library/src/main/scala/zio/elasticsearch/ElasticResult.scala
```diff
@@ -16,22 +16,67 @@
...
 
-import zio.prelude.ZValidation
+import zio.json.ast.Json
+import zio.json.ast.Json.Obj
+import zio.prelude.{Validation, ZValidation}
 import zio.schema.Schema
 import zio.{IO, Task, ZIO}
 
+sealed trait ElasticAggregationResult {
+  def getAggregations: Task[Map[String, ElasticAggregationResponse]]
```
drmarjanovic, line 26: Should we add a new line between two methods?

modules/library/src/main/scala/zio/elasticsearch/ElasticResult.scala
```diff
@@ -16,22 +16,67 @@
...
+  def getAggregation(name: String): Task[Option[ElasticAggregationResponse]]
+}
+
 sealed trait ElasticResult[F[_]] {
   def documentAs[A: Schema]: Task[F[A]]
 }
 
+sealed trait ElasticResultWithAggregation[F[_]] {
+  def documentAs[A: Schema]: Task[F[A]]
+  def getAggregations: Task[Map[String, ElasticAggregationResponse]]
```
drmarjanovic, line 36: Can we add new lines?

drmarjanovic (changes requested):

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregation.scala
```diff
@@ -0,0 +1,117 @@
...
+    TermsAggregation(
+      name = name,
+      field = field.toString ++ multiField.map("." ++ _).getOrElse(""),
+      subAggregations = Nil
+    )
+
+  def termsAggregation(name: String, field: String): ElasticAggregation[Terms] =
+    TermsAggregation(name = name, field = field, subAggregations = Nil)
+
+  private[elasticsearch] final case class MultipleAggregations(
```
drmarjanovic, line 62: Let's keep private methods at the bottom.

modules/library/src/main/scala/zio/elasticsearch/ElasticRequest.scala
```diff
@@ -37,6 +37,12 @@ sealed trait ElasticRequest[A]
 
 object ElasticRequest {
 
+  def aggregate(
```
drmarjanovic, line 40: Can we put this in one line?

modules/library/src/test/scala/zio/elasticsearch/utils/package.scala
```diff
@@ -18,10 +18,21 @@ package zio.elasticsearch
...
 
   final implicit class RichString(private val text: String) extends AnyVal {
     def toJson: Json = text.fromJson[Json].toOption.get
   }
+
+  final case class UserDocument(id: String, name: String, address: String, balance: Double, age: Int)
+
+  object UserDocument {
+
+    implicit val schema: Schema.CaseClass5[String, String, String, Double, Int, UserDocument] =
```
drmarjanovic, line 33: Do we need `Schema.CaseClass5[...]`?

modules/library/src/main/scala/zio/elasticsearch/package.scala
```diff
@@ -62,7 +62,21 @@ package object elasticsearch {
   def containsAny(name: String, params: List[String]): Boolean =
     params.exists(StringUtils.contains(name, _))
 
+  final implicit class ZIOAggregationOps[R](zio: RIO[R, ElasticAggregationResult]) {
+    def getAggregations: RIO[R, Map[String, ElasticAggregationResponse]] = zio.flatMap(_.getAggregations)
```
drmarjanovic, line 66: I would use resort this.

```
def getAggregation... 

def getAggregations...
```

modules/library/src/main/scala/zio/elasticsearch/SubAggregation.scala
```diff
@@ -0,0 +1,39 @@
...
+      aggregation: ElasticAggregation[EAT],
+      subAggregation: ElasticAggregation[_ <: SingleElasticAggregationType]
+    ): ElasticAggregation[EAT]
+  }
+
+  object WithSubAggregation {
+    implicit val termsWithSubAggregation: WithSubAggregation[Terms] =
+      (aggregation: ElasticAggregation[Terms], subAggregation: ElasticAggregation[_ <: SingleElasticAggregationType]) =>
+        aggregation match {
+          case a: TermsAggregation =>
```
drmarjanovic, line 35: Are we going to add other aggregations here? Maybe we should use `termsAgg: TermsAggregation`.

modules/library/src/main/scala/zio/elasticsearch/ElasticResult.scala
```diff
@@ -16,22 +16,73 @@
...
+import zio.json.ast.Json
+import zio.json.ast.Json.Obj
+import zio.prelude.{Validation, ZValidation}
 import zio.schema.Schema
 import zio.{IO, Task, ZIO}
 
+// todo better naming for traits and case classes!
+
+sealed trait ElasticAggregationResult {
+  def getAggregations: Task[Map[String, ElasticAggregationResponse]]
```
drmarjanovic, line 28: I would use sort - `getAggregation` and `getAggregations`.

modules/library/src/main/scala/zio/elasticsearch/ElasticResult.scala
```diff
@@ -47,3 +98,46 @@ final class SearchResult private[elasticsearch] (private val hits: List[Item]) e
...
+              .map(_.toMap)
+          case Left(e) =>
+            Left(DecodingException(s"Could not parse all documents successfully: $e)"))
+        }
+      case None =>
+        Right(Map.empty[String, BucketAggregationResponse])
+    }
+  }
+
+  def getAggregation(name: String): Task[Option[ElasticAggregationResponse]] = getAggregations.map(_.get(name))
```
drmarjanovic, line 142: Same here.

drmarjanovic (changes requested):

modules/library/src/main/scala/zio/elasticsearch/ElasticResult.scala
```diff
@@ -16,22 +16,73 @@
...
+              .map(_.toMap)
+          case Left(e) =>
+            Left(DecodingException(s"Could not parse all documents successfully: $e)"))
+        }
+      case None =>
+        Right(Map.empty[String, BucketAggregationResponse])
+    }
+  }
+
+  def getAggregation(key: String): Task[Option[ElasticAggregationResponse]] = getAggregations.map(_.get(key))
```
drmarjanovic, line 75: Also, I would use `aggregation` instead of `getAggregation`.

modules/library/src/test/scala/zio/elasticsearch/utils/package.scala
dbulaja98: Yes, maybe @mvelimir can provide some info about that?

modules/library/src/main/scala/zio/elasticsearch/HttpElasticExecutor.scala
```diff
@@ -42,32 +42,53 @@ private[elasticsearch] final class HttpElasticExecutor private (config: ElasticC
...
+      request
+        .post(uri"${config.uri}/${r.index}/$Search")
+        .response(asJson[ElasticQueryResponse])
+        .contentType(ApplicationJson)
+        .body(r.aggregation.toJson)
+    ).flatMap { response =>
+      response.code match {
+        case HttpOk =>
+          response.body.fold(
+            e => ZIO.fail(new ElasticException(s"Exception occurred: ${e.getMessage}")),
```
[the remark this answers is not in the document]
dbulaja98: We did it like this for executeSearch, so I repeated it here.

modules/library/src/main/scala/zio/elasticsearch/DocumentResult.scala
```diff
@@ -0,0 +1,226 @@
...
+        Right(None)
+    }
+  }
+
+  def aggregations: Task[Map[String, ElasticAggregationResponse]] = ZIO.fromEither {
+    aggregationsJson match {
+      case Some(aggregations) =>
+        Obj.decoder.decodeJson(aggregations.toString) match {
+          case Right(res) =>
+            Validation
```
drmarjanovic, line 87: Can we extract this to `private def`? We have the same/similar code for `aggregation`.

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregation.scala
```diff
@@ -0,0 +1,117 @@
...
+
+    def aggregations(aggregations: ElasticAggregation[_ <: SingleElasticAggregationType]*): MultipleAggregations =
+      self.copy(aggregations = self.aggregations ++ aggregations)
+
+    def paramsToJson: Json = Obj(aggregations.map { a =>
+      (a: @unchecked) match {
+        case TermsAggregation(name, field, subAggregations) =>
+          (
+            name,
+            Obj(List("terms" -> Obj("field" -> field.toJson)) ::: subAggregations.map { aggr =>
```
drmarjanovic, line 74: Can we use `agg` instead of `aggr`?

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregation.scala
```diff
@@ -0,0 +1,117 @@
...
+    TermsAggregation(name = name, field = field, subAggregations = Nil)
+
+  private[elasticsearch] final case class MultipleAggregations(
+    aggregations: List[ElasticAggregation[_ <: SingleElasticAggregationType]]
+  ) extends ElasticAggregation[Multiple] { self =>
+
+    def aggregations(aggregations: ElasticAggregation[_ <: SingleElasticAggregationType]*): MultipleAggregations =
+      self.copy(aggregations = self.aggregations ++ aggregations)
+
+    def paramsToJson: Json = Obj(aggregations.map { a =>
```
drmarjanovic, line 69: I would use `agg` instead of `a`.

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregation.scala
```diff
@@ -0,0 +1,117 @@
...
+
+  private[elasticsearch] final case class TermsAggregation(
+    name: String,
+    field: String,
+    subAggregations: List[ElasticAggregation[_ <: SingleElasticAggregationType]]
+  ) extends ElasticAggregation[Terms] {
+
+    def paramsToJson: Json = Obj(name -> paramsToJsonHelper(field, subAggregations))
+
+    def paramsToJsonHelper(
```
drmarjanovic, line 94: I was wondering if we can make this private.

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregation.scala
```diff
@@ -0,0 +1,117 @@
...
+      currField: String,
+      currSubAggs: List[ElasticAggregation[_ <: SingleElasticAggregationType]]
+    ): Obj =
+      if (currSubAggs.nonEmpty)
+        Obj(List("terms" -> Obj("field" -> currField.toJson)) :+ ("aggs" -> Obj(currSubAggs.map { aggr =>
```
drmarjanovic, line 99: Let's use `agg` instead of `aggr`.

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregation.scala
```diff
@@ -0,0 +1,117 @@
...
+          (aggr: @unchecked) match {
+            case a: TermsAggregation => (a.name, paramsToJsonHelper(a.field, a.subAggregations))
```
drmarjanovic, line 101: I would probably use `termsAgg` here.

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregation.scala
```diff
@@ -0,0 +1,117 @@
...
+          }
+        }: _*)): _*)
+      else
+        Obj(List("terms" -> Obj("field" -> currField.toJson)): _*)
+
```
drmarjanovic, line 106: Do we need this empty line?

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregation.scala
```diff
@@ -0,0 +1,117 @@
...
+  }
+}
+
+sealed trait ElasticAggregationType
+
+sealed trait SingleElasticAggregationType extends ElasticAggregationType
+
+object ElasticAggregationType {
+  sealed trait Terms    extends SingleElasticAggregationType
```
drmarjanovic, line 115: Let's put `Multiple` first, and then `Terms`.

modules/library/src/main/scala/zio/elasticsearch/OtherAggregation.scala
```diff
@@ -0,0 +1,52 @@
...
+package zio.elasticsearch
+
+import zio.elasticsearch.ElasticAggregation.{MultipleAggregations, TermsAggregation, multipleAggregations}
+import zio.elasticsearch.ElasticAggregationType.{Multiple, Terms}
+
+object OtherAggregation {
+
+  trait WithOtherAggregation[EAT <: ElasticAggregationType] {
+    def withAgg(
+      aggregation: ElasticAggregation[EAT],
```
drmarjanovic, line 26: Does it make sense to use `agg` and `otherAgg`?

drmarjanovic (changes requested):

modules/library/src/test/scala/zio/elasticsearch/utils/package.scala
mvelimir: We can't annotate it as `Schema[UserDocument`, but you could remove the type and make sure to put ```// scalafix:ok``` on that line.

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregation.scala
```diff
@@ -0,0 +1,115 @@
...
+
+  private[elasticsearch] final case class MultipleAggregations(
+    aggregations: List[ElasticAggregation[_ <: SingleElasticAggregationType]]
+  ) extends ElasticAggregation[Multiple] { self =>
+
+    def aggregations(aggregations: ElasticAggregation[_ <: SingleElasticAggregationType]*): MultipleAggregations =
+      self.copy(aggregations = self.aggregations ++ aggregations)
+
+    def paramsToJson: Json = Obj(aggregations.map { agg =>
+      (agg: @unchecked) match {
```
arnoldlacko, line 68: Why does it need to be unchecked?

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregation.scala
```diff
@@ -0,0 +1,115 @@
...
+        case TermsAggregation(name, field, subAggregations) =>
+          (
+            name,
+            Obj(List("terms" -> Obj("field" -> field.toJson)) ::: subAggregations.map { agg =>
```
arnoldlacko, line 72: This `:::` could be replaced by a `::` as we're only prepending one element.

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregation.scala
```diff
@@ -0,0 +1,115 @@
...
+package zio.elasticsearch
+
+import zio.elasticsearch.ElasticAggregationType.{Multiple, Terms}
+import zio.elasticsearch.ElasticPrimitive._
+import zio.elasticsearch.OtherAggregation.WithOtherAggregation
+import zio.elasticsearch.SubAggregation.WithSubAggregation
+import zio.json.ast.Json
+import zio.json.ast.Json.Obj
+
+sealed trait ElasticAggregation[EAT <: ElasticAggregationType] { self =>
```
arnoldlacko, line 26: Can we replace the phantom type-parameter with specialised traits as we've done in `ElasticRequest`?

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregation.scala
```diff
@@ -0,0 +1,115 @@
...
+    ): Obj =
+      if (currSubAggs.nonEmpty)
+        Obj(List("terms" -> Obj("field" -> currField.toJson)) :+ ("aggs" -> Obj(currSubAggs.map { agg =>
+          (agg: @unchecked) match {
+            case termsAgg: TermsAggregation =>
+              (termsAgg.name, paramsToJsonHelper(termsAgg.field, termsAgg.subAggregations))
+          }
+        }: _*)): _*)
+      else
+        Obj(List("terms" -> Obj("field" -> currField.toJson)): _*)
```
arnoldlacko, line 104: I think this might work without `List` and `: _*`

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregation.scala
```diff
@@ -0,0 +1,115 @@
...
+      currField: String,
+      currSubAggs: List[ElasticAggregation[_ <: SingleElasticAggregationType]]
+    ): Obj =
+      if (currSubAggs.nonEmpty)
+        Obj(List("terms" -> Obj("field" -> currField.toJson)) :+ ("aggs" -> Obj(currSubAggs.map { agg =>
+          (agg: @unchecked) match {
+            case termsAgg: TermsAggregation =>
+              (termsAgg.name, paramsToJsonHelper(termsAgg.field, termsAgg.subAggregations))
+          }
+        }: _*)): _*)
```
arnoldlacko, line 102: Perhaps we could just rely on the `vararg` syntax instead of creating a `List`?

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregationResponse.scala
```diff
@@ -0,0 +1,100 @@
...
+          field match {
+            case "key" =>
+              Some(field -> data.toString.replaceAll("\"", ""))
+            case "doc_count" =>
+              data.as[Int].toOption.map(field -> _)
+            case _ =>
+              val objFields = Obj.decoder.decodeJson(data.toString).toOption.get.fields.toMap
+
+              field match {
+                case str if str.contains("terms") =>
```
arnoldlacko, line 63: Should we use something more restrictive than `contains`? Could this be a problem if someone defines a custom aggregation like `mytermsagg`?

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregationResponse.scala
```diff
@@ -0,0 +1,100 @@
...
+                      buckets = objFields("buckets")
+                        .as[List[Json]]
+                        .toOption
+                        .get
+                        .map(bucket => TermsAggregationBucket.decoder.decodeJson(bucket.toJson).toOption)
+                        .map(_.get)
+                    )
+                  )
+                case _ =>
+                  throw new RuntimeException("Exception happened while decoding TermsAggregationBucket.")
```
arnoldlacko, line 77: Should we expose some more details?

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregationResponse.scala
```diff
@@ -0,0 +1,100 @@
...
+          }
+        }.toMap
+
+        val key      = allFields("key").asInstanceOf[String]
+        val docCount = allFields("doc_count").asInstanceOf[Int]
+        val subAggs = allFields.filter { case (field, _) =>
+          field != "key" && field != "doc_count"
+        }.map { case (field, data) =>
+          field match {
+            case str if str.contains("terms") =>
```
arnoldlacko, line 88: Same question as above about having a more restrictive option than `contains`

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregationResponse.scala
```diff
@@ -0,0 +1,100 @@
...
+              (field.split("#")(1), data.asInstanceOf[TermsAggregationResponse])
+            case _ =>
+              throw new RuntimeException("Exception happened while decoding TermsAggregationBucket.")
```
arnoldlacko, line 91: Perhaps we could expose some more information?

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregationResponse.scala
```diff
@@ -0,0 +1,100 @@
...
+          }
+        }
+
+        TermsAggregationBucket.apply(key, docCount, Option(subAggs).filter(_.nonEmpty))
+      case _ =>
+        throw new RuntimeException("Exception happened while decoding TermsAggregationBucket.")
```
arnoldlacko, line 97: Perhaps we could expose some more information?

modules/library/src/main/scala/zio/elasticsearch/ElasticRequest.scala
```diff
@@ -37,6 +37,9 @@ sealed trait ElasticRequest[A]
 
 object ElasticRequest {
 
+  def aggregate(index: IndexName, aggregation: ElasticAggregation[_]): Aggregation =
```
arnoldlacko, line 40: Shouldn't the return type should be the public abstract one?

modules/library/src/main/scala/zio/elasticsearch/ElasticRequest.scala
```diff
@@ -70,9 +73,24 @@ object ElasticRequest {
   def search(index: IndexName, query: ElasticQuery[_, _]): Search =
     Search(index = index, query = query, routing = None)
 
+  def searchWithAggregation(
+    index: IndexName,
+    query: ElasticQuery[_, _],
+    aggregation: ElasticAggregation[_]
+  ): SearchWithAggregation =
```
arnoldlacko, line 80: Shouldn't the return type should be the public abstract one?

modules/library/src/main/scala/zio/elasticsearch/ElasticResult.scala
```diff
@@ -16,34 +16,138 @@
...
+    aggregationsJson match {
+      case Some(aggregations) =>
+        Obj.decoder.decodeJson(aggregations.toString) match {
+          case Right(res) =>
+            Validation
+              .validateAll(
+                getUsefulPairs(res.fields.toList).map { case (field, data) =>
+                  ZValidation.fromEither(
+                    field match {
+                      case str if str.contains("terms") =>
```
arnoldlacko, line 126: Similar decoding was implemented under `ElasticAggregationResponse`, why do we need both?

modules/library/src/main/scala/zio/elasticsearch/ElasticResult.scala
```diff
@@ -16,34 +16,138 @@
...
+}
+
+private object ElasticResult {
+  def decodeAggregationsJson(
+    aggregationsJson: Option[Json],
+    name: Option[String]
+  ): Either[DecodingException, Map[String, TermsAggregationResponse]] = {
+
+    val decodingExceptionMessage = name match {
+      case Some(_) => "Could not parse aggregation successfully."
```
arnoldlacko, line 107: Should we include the name in the message?

modules/library/src/main/scala/zio/elasticsearch/HttpElasticExecutor.scala
```diff
@@ -255,7 +295,7 @@ private[elasticsearch] final class HttpElasticExecutor private (config: ElasticC
                 getQueryParams(List((Scroll, Some(ScrollDefaultDuration)), ("routing", s.routing)))
               )
             )
-            .response(asJson[ElasticQueryResponse])
+            .response(asJson[SearchWithAggsResponse])
```
arnoldlacko, line 298: In this case we're dealing with a search request with no aggregations. Is there a need to parse the response to a type that includes aggregations or could we easily separate the response types?

modules/library/src/main/scala/zio/elasticsearch/HttpElasticExecutor.scala
```diff
@@ -42,32 +42,53 @@ private[elasticsearch] final class HttpElasticExecutor private (config: ElasticC
...
 
   def stream(r: SearchRequest): Stream[Throwable, Item] =
     ZStream.paginateChunkZIO("") { s =>
-      if (s.isEmpty) executeGetByQueryWithScroll(r) else executeGetByScroll(s)
+      if (s.isEmpty) executeSearchWithScroll(r) else executeGetByScroll(s)
     }
 
   def streamAs[A: Schema](r: SearchRequest): Stream[Throwable, A] =
     ZStream
       .paginateChunkZIO("") { s =>
```
drmarjanovic, line 67: I would add pattern match here.

drmarjanovic (changes requested):

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregation.scala
```diff
@@ -0,0 +1,105 @@
...
+
+sealed trait HasOtherAgg {
+  def withAgg(agg: SingleElasticAggregation): MultipleAggregations
+}
+
+sealed trait SingleElasticAggregation extends ElasticAggregation
+
+object ElasticAggregation {
+  def multipleAggregations: Multiple =
+    Multiple.empty
```
mvelimir, line 43: Would it make sense to drop this method and instead construct `Multiple` with a method taking var args, like we've done with `BoolQuery`?

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregationResponse.scala
```diff
@@ -0,0 +1,93 @@
...
+                  )
+              }
+          }
+        }.toMap
+
+        val key      = allFields("key").asInstanceOf[String]
+        val docCount = allFields("doc_count").asInstanceOf[Int]
+        val subAggs = allFields.filter { case (field, _) =>
+          field != "key" && field != "doc_count"
+        }.map { case (field, data) =>
```
mvelimir, line 83: `filter` and `map` is just `collect`.

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregationResponse.scala
```diff
@@ -0,0 +1,93 @@
...
+          (field: @unchecked) match {
+            case str if str.contains("terms#") =>
```
mvelimir, line 85: This could just be an if statement.

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregationResponse.scala
```diff
@@ -0,0 +1,93 @@
...
+  docCount: Int,
+  subAggregations: Option[Map[String, ElasticAggregationResponse]] = None
+) extends ElasticAggregationBucket
+
+object TermsAggregationBucket {
+  implicit val decoder: JsonDecoder[TermsAggregationBucket] = (trace: List[JsonError], in: RetractReader) => {
+    val json = JsonDecoder[Json].unsafeDecode(trace, in)
+
+    (json: @unchecked) match {
+      case Obj(fields) =>
```
mvelimir, line 52: I think you could just decode this to `Obj` instead of `Json`, and avoid the need for the `@unchecked` annotation.

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregationResponse.scala
```diff
@@ -0,0 +1,93 @@
...
+          field match {
+            case "key" =>
+              Some(field -> data.toString.replaceAll("\"", ""))
+            case "doc_count" =>
+              data.as[Int].toOption.map(field -> _)
+            case _ =>
+              val objFields = Obj.decoder.decodeJson(data.toString).toOption.get.fields.toMap
+
+              (field: @unchecked) match {
+                case str if str.contains("terms#") =>
```
mvelimir, line 63: This could just be an if statement.

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregationResponse.scala
```diff
@@ -0,0 +1,93 @@
...
+final case class TermsAggregationBucket(
+  key: String,
+  @jsonField("doc_count")
+  docCount: Int,
+  subAggregations: Option[Map[String, ElasticAggregationResponse]] = None
+) extends ElasticAggregationBucket
+
+object TermsAggregationBucket {
+  implicit val decoder: JsonDecoder[TermsAggregationBucket] = (trace: List[JsonError], in: RetractReader) => {
+    val json = JsonDecoder[Json].unsafeDecode(trace, in)
```
mvelimir, line 49: Instead of doing it like this, maybe consider:
```scala
  implicit val decoder: JsonDecoder[TermsAggregationBucket] = JsonDecoder[Obj].mapOrFail(obj => ...)
```
I think it would make sense if the semantics were the same, which should be looked into.

mvelimir: Also, I've seen you use `Obj.decoder` in a few places, so either use that in this case too for consistency, or change the rest to `JsonDecoder[Obj]`.

modules/library/src/main/scala/zio/elasticsearch/ElasticResult.scala
```diff
@@ -16,34 +16,138 @@
...
+                  s"$decodingExceptionMessage: ${errors.mkString(",")})"
+                )
+              }
+              .map(_.toMap)
+          case Left(e) =>
+            Left(DecodingException(s"$decodingExceptionMessage: $e)"))
+        }
+      case None =>
+        Right(Map.empty[String, TermsAggregationResponse])
+    }
```
mvelimir, line 151: Instead of pattern matching on `Option`, you can use `fold`.

modules/library/src/main/scala/zio/elasticsearch/ElasticAggregationResponse.scala
```diff
@@ -0,0 +1,93 @@
...
+            case "doc_count" =>
+              data.as[Int].toOption.map(field -> _)
+            case _ =>
+              val objFields = Obj.decoder.decodeJson(data.toString).toOption.get.fields.toMap
+
+              (field: @unchecked) match {
+                case str if str.contains("terms#") =>
+                  Some(
+                    field -> TermsAggregationResponse(
+                      docErrorCount = objFields("doc_count_error_upper_bound").as[Int].toOption.get,
```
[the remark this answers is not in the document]
arnoldlacko: I think it makes sense to create a private `unsafeAs[A]`

drmarjanovic (approved):

PR #96, opened 2023-02-22
Pass routing parameter when executing delete by query

mvelimir:


modules/library/src/main/scala/zio/elasticsearch/Routing.scala
```diff
@@ -63,6 +72,17 @@ object Routing extends Newtype[String] {
...
     }
 
+    implicit val deleteByQueryWithRouting: WithRouting[DeleteByQuery] = new WithRouting[DeleteByQuery] {
+      def withRouting[A](
+        request: ElasticRequest[A, DeleteByQuery],
+        routing: Routing
+      ): ElasticRequest[A, DeleteByQuery] =
+        request match {
+          case Map(r, mapper)          => Map(withRouting(r, routing), mapper)
+          case r: DeleteByQueryRequest => r.copy(routing = Some(routing))
```
drmarjanovic, line 82: Should we put `Map` at the bottom? Not sure about consistency with existing code?

mvelimir: That is the order for all routing and refresh instances, so we would have to change it there too

dbulaja98 (approved):

modules/library/src/main/scala/zio/elasticsearch/Routing.scala
drmarjanovic: Mea culpa 🙋‍♂️

drmarjanovic (approved):

Issue #87, opened 2023-02-07
Replace WireMock with STTP


drmarjanovic:
### Description

Replace the current implementation of WireMock with STTP. You can refer to the documentation [here](https://sttp.softwaremill.com/en/latest/testing.html).

WireMock is used to confirm that HttpElasticExecutor is creating and executing expected HTTP requests.

kevchuang:
I'm interested to work on this one

drmarjanovic:
All yours! 🎉

drmarjanovic:
@kevchuang, any news here? 😄 Thanks in advance. 🙏🏻

kevchuang:
Hey @drmarjanovic ! Sorry I was a bit busy these past weeks, I'm gonna work on it next week 😄

drmarjanovic:
Thanks! 🙏🏻  No problem! We are looking forward to your PRs! 🤩

PR #103, opened 2023-03-05
Replace WireMock with STTP

kevchuang:
This PR addresses #87.


dbulaja98 (approved):

modules/library/src/test/scala/zio/elasticsearch/SttpBackendStubSpec.scala
```diff
@@ -0,0 +1,216 @@
...
+import sttp.client3.{Request, Response, StringBody}
+import sttp.model.{Method, StatusCode}
+import zio.test.ZIOSpecDefault
+import zio.{Task, TaskLayer, ZLayer}
+
+trait SttpBackendStubSpec extends ZIOSpecDefault {
+
+  case class StubMapping(request: Request[_, _] => Boolean, response: Response[Any])
+
+  final implicit class SttpBackendStubImplicits[T](sttpBackendStub: SttpBackendStub[Task, T]) {
```
drmarjanovic, line 14: Does it make sense to use `SttpBackendStubOps` convention? Also, let's use `[A]`, instead of `[T]`. I always feel like `T` is Java convention/leftover. 😄

modules/library/src/test/scala/zio/elasticsearch/SttpBackendStubSpec.scala
```diff
@@ -0,0 +1,216 @@
...
+    def addStubMapping(stubMapping: StubMapping): SttpBackendStub[Task, T] =
+      sttpBackendStub
+        .whenRequestMatches(stubMapping.request)
+        .thenRespond(stubMapping.response)
+  }
+
+  val index: IndexName = IndexName("repositories")
```
drmarjanovic, line 21: If `index`, `repo` and `url` can be made `private` we should do that.

modules/library/src/test/scala/zio/elasticsearch/SttpBackendStubSpec.scala
```diff
@@ -0,0 +1,216 @@
...
+
+  val repo: GitHubRepo =
+    GitHubRepo(id = Some("123"), organization = "lambdaworks.io", name = "LambdaWorks", stars = 10, forks = 10)
+
+  val url = "http://localhost:9200"
+
+  val bulkRequestStub: StubMapping = StubMapping(
```
drmarjanovic, line 28: Let's make all of those private if possible.

modules/library/src/test/scala/zio/elasticsearch/SttpBackendStubSpec.scala
```diff
@@ -0,0 +1,216 @@
...
+    createRequestWithGivenIdStub,
+    deleteByIdRequestStub,
+    deleteByQueryRequestStub,
+    deleteIndexRequestStub,
+    existsRequestStub,
+    getByIdRequestStub,
+    getByQueryRequestStub
+  )
+
+  val sttpBackendStubLayer: TaskLayer[SttpBackendStub[Task, Any]] = ZLayer.succeed(
```
drmarjanovic, line 209: This can be private as well, right?

drmarjanovic (commented):
Great job once again! I have several comments regarding the convention we use. Thanks 🙏🏻

modules/library/src/test/scala/zio/elasticsearch/SttpBackendStubSpec.scala
```diff
@@ -0,0 +1,216 @@
...
+import sttp.client3.httpclient.zio.HttpClientZioBackend
+import sttp.client3.testing.SttpBackendStub
+import sttp.client3.{Request, Response, StringBody}
+import sttp.model.{Method, StatusCode}
+import zio.test.ZIOSpecDefault
+import zio.{Task, TaskLayer, ZLayer}
+
+trait SttpBackendStubSpec extends ZIOSpecDefault {
+
+  case class StubMapping(request: Request[_, _] => Boolean, response: Response[Any])
```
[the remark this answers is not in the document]
kevchuang: Adding `final` gives me an warning during compilation: 
`The outer reference in this type test cannot be checked at run time.
final case class StubMapping(request: Request[_, _] => Boolean, response: Response[Any])`
Do you have any idea on how to get rid of this warning while keeping `final` ?

modules/library/src/test/scala/zio/elasticsearch/SttpBackendStubSpec.scala
kevchuang: It makes totally sense !

drmarjanovic (approved):

Issue #67, opened 2023-02-03
Support sort parameter


dbulaja98:
### Description: 

The sort parameter allows us to add one or more sorts on specific fields. It is defined on a per field level, with special field name for _score to sort by score

The `order` option can have the following values:
- asc (default for all sortings except the `_score`)
- desc (default for sorting on the `_score`)

The `mode` option is used for sorting by array or multi-valued fields, and it can have the following values:
- min (default for asc)
- max (default for desc)
- sum
- avg
- median

```
POST /_search
{
   "query" : {
      "term" : { "product" : "chocolate" }
   },
   "sort" : [
      {"price" : {"order" : "asc", "mode" : "avg"}}
   ]
}
```

Elasticsearch also supports: sorting by fields that are inside one or more nested objects, handling docs which are missing the sort field, and more.. https://www.elastic.co/guide/en/elasticsearch/reference/7.17/sort-search-results.html

PR #117, opened 2023-03-17
Support Sort parameter

dbulaja98:
Resolves https://github.com/lambdaworks/zio-elasticsearch/issues/67

Still open questions about:
1. `unmappedType` parameter - we do not know all possible values
2. `format` parameter - if we use this param on Elasticsearch field that is number, Elasticsearch will return error. Also if we enter random string instead of expected one, it will return error. (we do not know all possible values)
3. `_score` and `_doc` can not be combined with other params (Elasticsearch returns error)

### DIFF
--- a/modules/library/src/it/scala/zio/elasticsearch/HttpExecutorSpec.scala
+++ b/modules/library/src/it/scala/zio/elasticsearch/HttpExecutorSpec.scala
@@ -21,7 +21,7 @@ import zio.elasticsearch.ElasticAggregation.{multipleAggregations, termsAggregat
 import zio.elasticsearch.ElasticQuery._
 import zio.elasticsearch.Mode.Max
 import zio.elasticsearch.Order.Desc
-import zio.elasticsearch.SortBy.sortBy
+import zio.elasticsearch.Sorting.sortBy
 import zio.stream.{Sink, ZSink}
 import zio.test.Assertion._
 import zio.test.TestAspect._
@@ -156,7 +156,7 @@ object HttpExecutorSpec extends IntegrationSpec {
             ElasticExecutor.execute(ElasticRequest.createIndex(firstSearchIndex)),
             ElasticExecutor.execute(ElasticRequest.deleteIndex(firstSearchIndex)).orDie
           ),
-          test("search using match all query with multiple terms aggregations with desc sort on one field") {
+          test("search using match all query with multiple terms aggregations with descending sort on one field") {
             checkOnce(genDocumentId, genEmployee, genDocumentId, genEmployee) {
               (firstDocumentId, firstEmployee, secondDocumentId, secondEmployee) =>
                 val firstEmployeeWithFixedAge  = firstEmployee.copy(age = 25)
--- a/modules/library/src/it/scala/zio/elasticsearch/HttpExecutorSpec.scala
+++ b/modules/library/src/it/scala/zio/elasticsearch/HttpExecutorSpec.scala
@@ -19,11 +19,15 @@ package zio.elasticsearch
 import zio.Chunk
 import zio.elasticsearch.ElasticAggregation.{multipleAggregations, termsAggregation}
 import zio.elasticsearch.ElasticQuery._
+import zio.elasticsearch.Mode.Max
+import zio.elasticsearch.Order.Desc
+import zio.elasticsearch.SortBy.sortBy
 import zio.stream.{Sink, ZSink}
 import zio.test.Assertion._
 import zio.test.TestAspect._
 import zio.test._
 
+import java.time.LocalDate
 import scala.util.Random
 
 object HttpExecutorSpec extends IntegrationSpec {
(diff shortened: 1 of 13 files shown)


modules/library/src/main/scala/zio/elasticsearch/ElasticRequest.scala
```diff
@@ -301,21 +305,31 @@ object ElasticRequest {
       self.copy(routing = Some(value))
   }
 
-  sealed trait SearchRequest extends ElasticRequest[SearchResult]
+  sealed trait SearchRequest extends ElasticRequest[SearchResult] with WithSort[SearchRequest]
 
   private[elasticsearch] final case class Search(
     index: IndexName,
     query: ElasticQuery[_],
+    sortBy: Option[List[SortBy]],
```
drmarjanovic, line 313: Can we use `List[SortBy]` only here? It's empty when `Nil`.

modules/library/src/main/scala/zio/elasticsearch/ElasticRequest.scala
```diff
@@ -301,21 +305,31 @@ object ElasticRequest {
...
     routing: Option[Routing]
-  ) extends SearchRequest
+  ) extends SearchRequest { self =>
+    def sortBy(sorts: SortBy*): SearchRequest =
+      self.copy(sortBy = Some(sorts.toList))
```
drmarjanovic, line 317: Does it make sense to append to existing sorts? Also, does it make sense to use `Set`?

drmarjanovic (changes requested):
Can we add tests with `Class.field` support instead of plain strings?

modules/library/src/main/scala/zio/elasticsearch/HttpElasticExecutor.scala
```diff
@@ -308,13 +308,19 @@ private[elasticsearch] final class HttpElasticExecutor private (esConfig: Elasti
       }
     }
 
-  private def executeSearch(r: Search): Task[SearchResult] =
+  private def executeSearch(r: Search): Task[SearchResult] = {
+    val body = r.sortBy match {
+      case sorts if sorts.nonEmpty =>
+        Obj(List("query" -> r.query.paramsToJson(None), "sort" -> Arr(sorts.toList.map(_.paramsToJson): _*)): _*)
+      case _ => r.query.toJson
```
drmarjanovic, line 315: Move `r.query.toJson` to a new line.

modules/library/src/main/scala/zio/elasticsearch/HttpElasticExecutor.scala
```diff
@@ -388,13 +395,27 @@ private[elasticsearch] final class HttpElasticExecutor private (esConfig: Elasti
     }
   }
 
-  private def executeSearchWithAggregation(r: SearchWithAggregation): Task[SearchWithAggregationsResult] =
+  private def executeSearchWithAggregation(r: SearchWithAggregation): Task[SearchWithAggregationsResult] = {
+    val body = r.sortBy match {
+      case sorts if sorts.nonEmpty =>
+        Obj(
+          List(
+            "query" -> r.query.paramsToJson(None),
```
drmarjanovic, line 403: Can you name `None` param?

modules/library/src/main/scala/zio/elasticsearch/HttpElasticExecutor.scala
```diff
@@ -388,13 +395,27 @@ private[elasticsearch] final class HttpElasticExecutor private (esConfig: Elasti
...
+            "sort"  -> Arr(sorts.toList.map(_.paramsToJson): _*),
+            "aggs"  -> r.aggregation.paramsToJson
+          ): _*
+        )
+      case _ =>
+        Obj(List("query" -> r.query.paramsToJson(None), "aggs" -> r.aggregation.paramsToJson): _*)
```
drmarjanovic, line 409: Same here. And you can use same format: 

```
Obj(
  List(
    query -> 
    aggs -> 
  )
)
```

modules/library/src/it/scala/zio/elasticsearch/HttpExecutorSpec.scala
```diff
@@ -152,6 +156,48 @@ object HttpExecutorSpec extends IntegrationSpec {
             ElasticExecutor.execute(ElasticRequest.createIndex(firstSearchIndex)),
             ElasticExecutor.execute(ElasticRequest.deleteIndex(firstSearchIndex)).orDie
           ),
+          test("search using match all query with multiple terms aggregations with desc sort on one field") {
```
drmarjanovic, line 159: Let's use `descending` instead of `desc`.

drmarjanovic (changes requested):

modules/library/src/main/scala/zio/elasticsearch/SortBy.scala
```diff
@@ -0,0 +1,189 @@
...
+      field = field,
+      format = None,
+      mode = None,
+      missing = None,
+      numericType = None,
+      order = None,
+      unmappedType = None
+    )
+
+  private[elasticsearch] final case class SortByData(
```
arnoldlacko, line 148: I wonder what better names we could find for `SortByData extends SortBy`, one that comes to mind is `SortConfig extends Sorting` 🤔

modules/library/src/main/scala/zio/elasticsearch/ElasticResult.scala
```diff
@@ -79,7 +71,8 @@ final class SearchResult private[elasticsearch] (private val hits: List[Item]) e
 final class SearchWithAggregationsResult private[elasticsearch] (
   private val hits: List[Item],
   private val aggregationsJson: Option[Json]
-) extends DocumentsWithAggregationsResult {
+) extends DocumentResult[List]
+    with AggregationsResult {
```
drmarjanovic, line 75: Can we have `extends DocumentResult[List] with AggregationsResult`? In one line?

modules/library/src/it/scala/zio/elasticsearch/HttpExecutorSpec.scala
```diff
@@ -444,6 +490,84 @@ object HttpExecutorSpec extends IntegrationSpec {
...
+                           .documentAs[EmployeeDocument]
+                } yield assert(res)(
+                  equalTo(List(secondCustomerWithFixedAge, firstCustomerWithFixedAge))
+                )
+            }
+          } @@ around(
+            ElasticExecutor.execute(ElasticRequest.createIndex(firstSearchIndex)),
+            ElasticExecutor.execute(ElasticRequest.deleteIndex(firstSearchIndex)).orDie
+          ),
+          test("search for document sorted descending with 'max' mode by one field using matchAll query") {
```
drmarjanovic, line 534: Let's add at least one `ascending`.

drmarjanovic (changes requested):

modules/library/src/main/scala/zio/elasticsearch/ElasticResult.scala
dbulaja98: `Scalafmt` does not allow it.

drmarjanovic (approved):

PR #123, opened 2023-03-20
(api): Support count

### LINKED ISSUE #122: Support Count API
### Description: 

Gets the number of matches for a search query.

```
GET /my-index-000001/_count?q=user:kimchy

GET /my-index-000001/_count
{
  "query" : {
    "term" : { "user.id" : "kimchy" }
  }
}

```

https://www.elastic.co/guide/en/elasticsearch/reference/7.17/search-count.html#search-count-api-path-params

dbulaja98:
Closes #122


modules/library/src/main/scala/zio/elasticsearch/HttpElasticExecutor.scala
```diff
@@ -120,6 +121,36 @@ private[elasticsearch] final class HttpElasticExecutor private (esConfig: Elasti
     }
   }
 
+  private def executeCount(r: Count): Task[Int] = {
+    val req = r.query match {
+      case Some(query) =>
+        request
```
drmarjanovic, line 127: Let's extract everything common, and apply `body` if `query` is not empty.

modules/library/src/main/scala/zio/elasticsearch/HttpElasticExecutor.scala
```diff
@@ -120,6 +121,36 @@ private[elasticsearch] final class HttpElasticExecutor private (esConfig: Elasti
...
+          .get(uri"${esConfig.uri}/${r.index}/$Count".withParams(getQueryParams(List(("routing", r.routing)))))
+          .contentType(ApplicationJson)
+          .response(asJson[ElasticCountResponse])
+    }
+
+    sendRequestWithCustomResponse(req).flatMap { response =>
+      response.code match {
+        case HttpOk =>
+          response.body
+            .map(res => res.count)
```
drmarjanovic, line 143: Can we use `_.count` here?

drmarjanovic: Is `count` `Int` or `Long` value?

drmarjanovic (changes requested):

modules/library/src/main/scala/zio/elasticsearch/HttpElasticExecutor.scala
dbulaja98: It is not specified, I suppose it can be both.

drmarjanovic (approved):

PR #130, opened 2023-03-24
Cross-compile to Scala 3

### LINKED ISSUE #90: Cross build to Scala 3
Are there any plans to target Scala 3? Is there anything standing in the way currently? I would be willing to help with the refactor if there is interest.

ioleo:
This PR sets up cross-compilation to Scala 3.

Closes #90.

### DIFF
--- a/project/BuildHelper.scala
+++ b/project/BuildHelper.scala
@@ -4,17 +4,17 @@ import scalafix.sbt.ScalafixPlugin.autoImport._
 
 object BuildHelper {
 
-  val Scala212: String   = "2.12.17"
-  val Scala213: String   = "2.13.10"
-  val ScalaDotty: String = "3.2.2"
+  val Scala212: String = "2.12.17"
+  val Scala213: String = "2.13.10"
+  val Scala3: String   = "3.2.2"
 
   def stdSettings(prjName: String) =
     List(
       name                     := s"$prjName",
-      crossScalaVersions       := List(Scala212, Scala213, ScalaDotty),
+      crossScalaVersions       := List(Scala212, Scala213, Scala3),
       ThisBuild / scalaVersion := Scala213,
       scalacOptions            := stdOptions ++ extraOptions(scalaVersion.value),
-      semanticdbEnabled        := scalaVersion.value != ScalaDotty, // enable SemanticDB
+      semanticdbEnabled        := scalaVersion.value != Scala3, // enable SemanticDB
       semanticdbOptions += "-P:semanticdb:synthetics:on",
       semanticdbVersion                                          := scalafixSemanticdb.revision,
       ThisBuild / scalafixScalaBinaryVersion                     := CrossVersion.binaryScalaVersion(scalaVersion.value),
@@ -24,7 +24,7 @@ object BuildHelper {
       autoAPIMappings := true
     )
 
-  private val stdOptions = Seq(
+  private val stdOptions = List(
     "-deprecation",
     "-encoding",
     "UTF-8",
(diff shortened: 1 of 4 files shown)


project/BuildHelper.scala
```diff
@@ -4,16 +4,17 @@ import scalafix.sbt.ScalafixPlugin.autoImport._
 
 object BuildHelper {
 
-  val Scala212: String = "2.12.17"
-  val Scala213: String = "2.13.10"
+  val Scala212: String   = "2.12.17"
+  val Scala213: String   = "2.13.10"
+  val ScalaDotty: String = "3.2.2"
```
drmarjanovic, line 9: Can we use `Scala3` instead of `ScalaDotty`?

project/BuildHelper.scala
```diff
@@ -23,10 +24,53 @@ object BuildHelper {
       autoAPIMappings := true
     )
 
-  private val extraOptions =
-    List("-Ywarn-unused")
+  private val stdOptions = Seq(
```
drmarjanovic, line 27: We prefer using `List` instead of `Seq`.

project/BuildHelper.scala
```diff
@@ -23,10 +24,53 @@ object BuildHelper {
...
+    "-Yrangepos",
+    "-Xlint:_,-missing-interpolator,-type-parameter-shadow",
+    "-Ywarn-numeric-widen",
+    "-Ywarn-value-discard"
+  )
 
+  def extraOptions(scalaVersion: String) =
+    CrossVersion.partialVersion(scalaVersion) match {
+      case Some((3, 2)) =>
+        Seq(
```
drmarjanovic, line 48: We should use `List` instead of `Seq` everywhere.

project/BuildHelper.scala
```diff
@@ -23,10 +24,53 @@ object BuildHelper {
...
+          "-Ywarn-infer-any",
+          "-Ywarn-nullary-override",
+          "-Ywarn-nullary-unit",
+          "-Ywarn-unused:params,-implicits",
+          "-Xfuture",
+          "-Xsource:2.13",
+          "-Xmax-classfile-name",
+          "242"
+        )
+      case _ => Seq()
```
drmarjanovic, line 74: Can we use `Nil` instead of `Seq()`?

drmarjanovic (commented):
Hi @ioleo, great work overall. I'm thrilled because we are adding Scala 3 support!

Can you add Scala 3 to our [ci.yml](https://github.com/lambdaworks/zio-elasticsearch/blob/main/.github/workflows/ci.yml) to verify it through tests as well?

Once again, thank you and we are looking forward to more contributions from you!

drmarjanovic (commented):
@ioleo, thanks for the quick response!

Can you just run `sbt prepare` before pushing? Seems our linter is failing! Thanks 🙏🏻

ioleo:
Hmm, I don't get the error here. `assert` and `assertCustom` are members of `NewType[String]`, so why would Scala 3 not find those?

ioleo:
@drmarjanovic ZIO Prelude 3.x is missing `assert` and `assertCustom` as they were backed by Scala 2 macro.
https://github.com/zio/zio-prelude/blob/series/2.x/core/shared/src/main/scala-2/zio/prelude/Newtype.scala#L258

Can we replace those calls (to `assert` and `assertCustom`)? Maybe more boilerplatish, but not relying on macros?
I haven't used ZIO Prelude, so I'm not sure what these do and why do you need them here in the codebase.

ioleo:
Continued in #131

PR #142, opened 2023-04-03
(api): Support `from` and `size` pagination parameters

### LINKED ISSUE #52: Support 'size' and 'from' pagination parameters
### Description:

`size` and `from` parameters are used to page through a larger set of search results. By default, Elasticsearch returns the top 10 matching hits, which is often not enough. 
The `from` parameter defines the number of top hits to skip (default is 0). The size parameter is the maximum number of hits to return.

_**NOTE:** By default, you cannot use `from` and `size` to page through more than 10,000 hits._

dbulaja98:
Resolves #52

### DIFF
--- a/modules/library/src/it/scala/zio/elasticsearch/HttpExecutorSpec.scala
+++ b/modules/library/src/it/scala/zio/elasticsearch/HttpExecutorSpec.scala
@@ -138,7 +138,7 @@ object HttpExecutorSpec extends IntegrationSpec {
           )
         ),
         suite("search with aggregation")(
-          test("search using match all query with multiple terms aggregations") {
+          test("search for first result using match all query with multiple terms aggregations") {
             checkOnce(genDocumentId, genTestDocument, genDocumentId, genTestDocument) {
               (firstDocumentId, firstDocument, secondDocumentId, secondDocument) =>
                 for {
@@ -164,10 +164,12 @@ object HttpExecutorSpec extends IntegrationSpec {
                                query = query,
                                aggregation = aggregation
                              )
+                             .from(0)
+                             .size(1)
                          )
                   docs <- res.documentAs[TestDocument]
                   aggs <- res.aggregations
-                } yield assert(docs)(isNonEmpty) && assert(aggs)(isNonEmpty)
+                } yield assert(docs.length)(equalTo(1)) && assert(aggs)(isNonEmpty)
             }
           } @@ around(
             Executor.execute(ElasticRequest.createIndex(firstSearchIndex)),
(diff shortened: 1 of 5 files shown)


modules/library/src/main/scala/zio/elasticsearch/executor/HttpExecutor.scala
```diff
@@ -383,7 +370,10 @@ private[elasticsearch] final class HttpExecutor private (esConfig: ElasticConfig
       )
     val defaultSortField = Json.Obj("sort" -> Json.Arr(Json.Str(ShardDoc)))
     val searchAfterJson  = searchAfter.map(sa => Json.Obj("search_after" -> sa))
-    val requestBody      = r.query.toJson merge pointInTimeJson merge defaultSortField
+    val sortsJson        = Obj("sort" -> Arr(r.sortBy.toList.map(_.paramsToJson): _*))
+    val requestBody =
+      if (r.sortBy.isEmpty) r.query.toJson merge pointInTimeJson merge defaultSortField
```
drmarjanovic, line 375: Let's use parenthesis.

modules/library/src/main/scala/zio/elasticsearch/ElasticRequest.scala
```diff
@@ -308,44 +319,112 @@ object ElasticRequest {
...
       self.copy(routing = Some(value))
 
+    def size(value: Int): SearchRequest =
+      self.copy(size = Some(value))
+
     def sortBy(sorts: Sort*): SearchRequest =
       self.copy(sortBy = sortBy ++ sorts.toSet)
+
+    def toJson: Json = {
+      val startJson = (self.from, self.size) match {
```
drmarjanovic, line 361: Let's use `baseJson` instead of `startJson`.

modules/library/src/main/scala/zio/elasticsearch/ElasticRequest.scala
```diff
@@ -308,44 +319,112 @@ object ElasticRequest {
...
       self.copy(routing = Some(value))
 
+    def size(value: Int): SearchAndAggregateRequest =
+      self.copy(size = Some(value))
+
     def sortBy(sorts: Sort*): SearchAndAggregateRequest =
       self.copy(sortBy = sortBy ++ sorts.toSet)
+
+    def toJson: Json = {
+      val startJson = (self.from, self.size) match {
```
drmarjanovic, line 410: Let's use `baseJson` instead of `startJson`.

drmarjanovic (changes requested):

modules/library/src/it/scala/zio/elasticsearch/HttpExecutorSpec.scala
```diff
@@ -424,9 +426,16 @@ object HttpExecutorSpec extends IntegrationSpec {
                            .upsert[TestDocument](firstSearchIndex, secondDocumentId, secondDocument)
                            .refreshTrue
```
markaya, line 427: You can remove this refresh and just leave it for last call before search.

markaya (changes requested):

drmarjanovic (approved):

modules/library/src/it/scala/zio/elasticsearch/HttpExecutorSpec.scala
dbulaja98: It is now.

drmarjanovic (approved):

PR #143, opened 2023-04-03
(api): Support Elasticsearch authorization

dbulaja98:


modules/library/src/main/scala/zio/elasticsearch/executor/HttpExecutor.scala
```diff
@@ -362,6 +368,8 @@ private[elasticsearch] final class HttpExecutor private (esConfig: ElasticConfig
             e => ZIO.fail(new ElasticException(s"Exception occurred: ${e.getMessage}")),
             value => ZIO.succeed(new SearchResult(value.results.map(Item.apply)))
           )
+        case HttpUnauthorized =>
+          ZIO.fail(UnauthorizedException("Unauthorized action!"))
```
drmarjanovic, line 372: Can we add a better message? 😄

modules/library/src/main/scala/zio/elasticsearch/executor/HttpExecutor.scala
```diff
@@ -362,6 +368,8 @@ private[elasticsearch] final class HttpExecutor private (esConfig: ElasticConfig
...
         case _ =>
```
drmarjanovic, line 373: It seems it's possible to get a `403` error as well. Check [here](https://www.elastic.co/guide/en/elasticsearch/reference/current/anonymous-access.html) for more info.

drmarjanovic (changes requested):

modules/library/src/main/scala/zio/elasticsearch/ElasticConfig.scala
```diff
@@ -17,11 +17,12 @@
...
 
 import sttp.model.Uri
+import zio.elasticsearch.executor.ElasticCredentials
 
-final case class ElasticConfig(host: String, port: Int) {
+final case class ElasticConfig(host: String, port: Int, credentials: Option[ElasticCredentials]) {
   lazy val uri: Uri = Uri(host, port)
 }
 
 object ElasticConfig {
```
drmarjanovic, line 26: Let's provide apply with `host` and `port` only.

modules/example/src/main/scala/example/Main.scala
```diff
@@ -32,7 +32,7 @@ import scala.util.Using
 object Main extends ZIOAppDefault {
 
   override def run: Task[ExitCode] = {
-    val elasticConfigLive = ZLayer(getConfig[ElasticsearchConfig].map(es => ElasticConfig(es.host, es.port)))
+    val elasticConfigLive = ZLayer(getConfig[ElasticsearchConfig].map(es => ElasticConfig(es.host, es.port, None)))
```
drmarjanovic, line 35: This should be unchanged.

modules/library/src/main/scala/zio/elasticsearch/ElasticConfig.scala
```diff
@@ -17,11 +17,12 @@
...
-  lazy val Default: ElasticConfig = ElasticConfig("localhost", 9200)
+  lazy val Default: ElasticConfig = ElasticConfig("localhost", 9200, None)
```
drmarjanovic, line 27: This should be unchanged.

modules/library/src/main/scala/zio/elasticsearch/executor/HttpExecutor.scala
```diff
@@ -489,25 +497,36 @@ private[elasticsearch] final class HttpExecutor private (esConfig: ElasticConfig
...
+      case HttpUnauthorized | HttpForbidden =>
+        UnauthorizedException
+      case _ =>
+        new ElasticException(
+          s"Unexpected response from Elasticsearch. Response body: ${response.body.fold(body => body, _ => "")}"
+        )
+    }
 
-  private def createElasticExceptionFromCustomResponse[A](
+  private def handleFailuresFromCustomResponse[A](
```
drmarjanovic, line 518: Let's have only `response` here. So you can match `response.code` instead of `errorCode`.

drmarjanovic (changes requested):

modules/library/src/main/scala/zio/elasticsearch/executor/HttpExecutor.scala
dbulaja98: Makes sense!

drmarjanovic (approved):

PR #147, opened 2023-04-06
Support last sort value and search after request

markaya:


modules/library/src/main/scala/zio/elasticsearch/result/ElasticResult.scala
```diff
@@ -65,6 +67,8 @@ final class SearchResult private[elasticsearch] (private val hits: List[Item]) e
     }
 
   def items: UIO[List[Item]] = ZIO.succeed(hits)
+
+  def lastSortValue(): ZIO[Any, Nothing, Option[Json]] = ZIO.succeed(lastSort)
```
drmarjanovic, line 71: Let's use `UIO` here. Also, it can be `lazy val`, right? Also, `items` can be `lazy val`.

drmarjanovic (changes requested):

modules/library/src/main/scala/zio/elasticsearch/ElasticRequest.scala
```diff
@@ -404,6 +415,7 @@ object ElasticRequest {
     from: Option[Int],
     highlights: Option[Highlights],
     routing: Option[Routing],
+    searchAfter: Option[Json],
     size: Option[Int]
   ) extends SearchAndAggregateRequest { self =>
     def from(value: Int): SearchAndAggregateRequest =
```
dbulaja98, line 421:
```
 def highlights(value: Highlights): SearchAndAggregateRequest =
      self.copy(highlights = Some(value))
```

Please provide this method in `sealed trait SearchAndAggregateRequest`
    def highlights(value: Highlights): SearchAndAggregateRequest

dbulaja98 (changes requested):

drmarjanovic (approved):

PR #163, opened 2023-04-13
(dsl): Add `withSuffix` method on `Field`

mvelimir:


drmarjanovic (commented):
What about adding `Field.keyword` method at the API level? 

Overall, a great improvement!

PR #184, opened 2023-04-25
Support `minimumShouldMatch` on BoolQuery

drmarjanovic:
### Description:

- Added `minimumShouldMatch` support for `BoolQuery`.
- Made the `paramsToJson` and `toJson` methods as package private.
- Formatted code.
- Adapted tests to check whether `minimumShouldMatch` is successfully set.

### DIFF
--- a/modules/library/src/main/scala/zio/elasticsearch/query/Queries.scala
+++ b/modules/library/src/main/scala/zio/elasticsearch/query/Queries.scala
@@ -24,9 +24,9 @@ import zio.schema.Schema
 import scala.annotation.unused
 
 sealed trait ElasticQuery[-S] { self =>
-  def paramsToJson(fieldPath: Option[String]): Json
+  private[elasticsearch] def paramsToJson(fieldPath: Option[String]): Json
 
-  final def toJson: Obj =
+  private[elasticsearch] final def toJson: Obj =
     Obj("query" -> self.paramsToJson(None))
 }
 
@@ -87,7 +87,8 @@ private[elasticsearch] final case class Bool[S](
         if (must.nonEmpty) Some("must" -> Arr(must.map(_.paramsToJson(fieldPath)): _*)) else None,
         if (mustNot.nonEmpty) Some("must_not" -> Arr(mustNot.map(_.paramsToJson(fieldPath)): _*)) else None,
         if (should.nonEmpty) Some("should" -> Arr(should.map(_.paramsToJson(fieldPath)): _*)) else None,
-        boost.map("boost" -> Num(_))
+        boost.map("boost" -> Num(_)),
+        minimumShouldMatch.map("minimum_should_match" -> Num(_))
       ).collect { case Some(obj) => obj }
 
     Obj("bool" -> Obj(boolFields: _*))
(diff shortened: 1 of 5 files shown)


modules/library/src/main/scala/zio/elasticsearch/query/Queries.scala
```diff
@@ -64,6 +65,9 @@ private[elasticsearch] final case class Bool[S](
   def filter(queries: ElasticQuery[Any]*): BoolQuery[S] =
     self.copy(filter = filter ++ queries)
 
+  def minimumShouldMatch(value: Int): BoolQuery[S] =
+    self.copy(minimumShouldMatch = Some(value))
+
```
markaya, line 70: Move this after `mustNot`?

modules/library/src/main/scala/zio/elasticsearch/query/package.scala
```diff
@@ -120,6 +120,21 @@ package object query {
     def innerHits(innerHits: InnerHits): Q
   }
 
+  private[elasticsearch] trait HasMinimumShouldMatch[Q <: HasBoost[Q]] {
```
markaya, line 123:
```suggestion
  private[elasticsearch] trait HasMinimumShouldMatch[Q <: HasMinimumShouldMatch[Q]] {
```

markaya (commented):
Maybe we should add test to check if it is written in file successfully?

Also it would be nice to have one integration test.

modules/library/src/main/scala/zio/elasticsearch/query/Queries.scala
drmarjanovic: Sorted alphabetically.

modules/library/src/main/scala/zio/elasticsearch/query/package.scala
```diff
@@ -120,6 +120,21 @@ package object query {
...
+
+    /**
+     * Sets the `minimumShouldMatch` parameter for this [[ElasticQuery]]. The `minimumShouldMatch` value is the number
+     * of should clauses returned documents must match. If the [[zio.elasticsearch.query.BoolQuery]] includes at least
+     * one `should` clause and no `must`/`filter` clauses, the default value is 1. Otherwise, the default value is 0.
+     *
+     * @param value
+     *   a number to set `minimumShouldMatch` parameter to
+     * @return
+     *   a new instance of the [[ElasticQuery]] with the `minimumShouldMatch` value set.
```
dbulaja98, line 133: If this applies only on `BoolQuery`, maybe write it here instead of `ElasticQuery`.

drmarjanovic: Not sure if there are any other places to be supported.

drmarjanovic: At least [match bool prefix](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/query-dsl-match-bool-prefix-query.html) query supports this. So, let's have it in a separate trait.

markaya (approved):

arnoldlacko (approved):

PR #185, opened 2023-04-25
Add support to access search response fields

markaya:
Summary:
- Added full response to SearchResult class so we can support accessing different fields from response. (should discuss further)
- Added sort field to Item so we can access sort field of each individual document returned

### DIFF
--- a/modules/library/src/main/scala/zio/elasticsearch/executor/response/SearchWithAggregationsResponse.scala
+++ b/modules/library/src/main/scala/zio/elasticsearch/executor/response/SearchWithAggregationsResponse.scala
@@ -55,9 +55,8 @@ private[elasticsearch] final case class SearchWithAggregationsResponse(
       )
       .toEitherWith(_.mkString(", "))
 
-  lazy val results: List[Json] = hits.hits.map(_.source)
-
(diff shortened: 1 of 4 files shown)


modules/library/src/main/scala/zio/elasticsearch/executor/response/SearchWithAggregationsResponse.scala
```diff
@@ -55,9 +55,8 @@ private[elasticsearch] final case class SearchWithAggregationsResponse(
       )
       .toEitherWith(_.mkString(", "))
 
-  lazy val results: List[Json] = hits.hits.map(_.source)
-
-  lazy val resultsWithHighlights: List[(Json, Option[Json])] = hits.hits.map(h => (h.source, h.highlight))
+  lazy val resultsWithHighlightsAndSort: List[(Json, Option[Json], Option[Json])] =
```
arnoldlacko, line 58: We should think about a dedicated type for the tuple `(Json, Option[Json], Option[Json])` so we wouldn't need to communicate the meaning of each element through a complicated value name.

markaya: Yes definetly. It would be much easier to understand if we had case class for this. I will implement it, just not sure about naming of such class.

PR #193, opened 2023-04-28
Restrict `sort` and `order` methods

dbulaja98:


modules/library/src/main/scala/zio/elasticsearch/request/options/HasSort.scala
```diff
@@ -24,10 +24,12 @@ private[elasticsearch] trait HasSort[R <: HasSort[R]] {
    * Sets the sorting criteria for the [[zio.elasticsearch.ElasticRequest.SearchRequest]] or the
    * [[zio.elasticsearch.ElasticRequest.SearchAndAggregateRequest]].
    *
+   * @param sort
+   *   one [[zio.elasticsearch.query.sort.Sort]] object that define the sorting criteria
```
drmarjanovic, line 28: `required ... object that define`

modules/library/src/main/scala/zio/elasticsearch/request/options/HasSort.scala
```diff
@@ -24,10 +24,12 @@ private[elasticsearch] trait HasSort[R <: HasSort[R]] {
...
    * @param sorts
-   *   one or more [[zio.elasticsearch.query.sort.Sort]] objects that define the sorting criteria
+   *   zero or more [[zio.elasticsearch.query.sort.Sort]] objects that define the sorting criteria
```
drmarjanovic, line 30: ...rest of the Sort objects that define...

modules/library/src/main/scala/zio/elasticsearch/ElasticRequest.scala
```diff
@@ -600,8 +600,8 @@ object ElasticRequest {
     def size(value: Int): SearchRequest =
       self.copy(size = Some(value))
 
-    def sort(sorts: Sort*): SearchRequest =
-      self.copy(sortBy = sortBy ++ sorts.toSet)
+    def sort(sort: Sort, sorts: Sort*): SearchRequest =
+      self.copy(sortBy = (sortBy :+ sort) ++ sorts.toList)
```
arnoldlacko, line 604: If we change `(sortBy :+ sort) ++ sorts.toList` to `sortBy ++ (sort :: sorts.toList)` then `sortBy` only gets copied once instead of twice.
Also, should we use Chunk instead?

modules/library/src/main/scala/zio/elasticsearch/ElasticRequest.scala
```diff
@@ -656,8 +656,8 @@ object ElasticRequest {
     def searchAfter(value: Json): SearchAndAggregateRequest =
       self.copy(searchAfter = Some(value))
 
-    def sort(sorts: Sort*): SearchAndAggregateRequest =
-      self.copy(sortBy = sortBy ++ sorts.toSet)
+    def sort(sort: Sort, sorts: Sort*): SearchAndAggregateRequest =
+      self.copy(sortBy = (sortBy :+ sort) ++ sorts.toList)
```
arnoldlacko, line 660: If we change (sortBy :+ sort) ++ sorts.toList to sortBy ++ (sort :: sorts.toList) then sortBy only gets copied once instead of twice.

On a separate note: should we use Chunk instead?

modules/library/src/main/scala/zio/elasticsearch/aggregation/Aggregations.scala
```diff
@@ -68,12 +68,12 @@ sealed trait TermsAggregation
...
   name: String,
   field: String,
-  order: Set[AggregationOrder],
+  order: List[AggregationOrder],
   subAggregations: List[SingleElasticAggregation],
   size: Option[Int]
 ) extends TermsAggregation { self =>
   def orderBy(order: AggregationOrder, orders: AggregationOrder*): TermsAggregation =
-    self.copy(order = self.order + order ++ orders.toSet)
+    self.copy(order = (self.order :+ order) ++ orders)
```
arnoldlacko, line 76: Same as above:
- self.order is copied twice this way instead of just once
- should we use chunk?

drmarjanovic (approved):

PR #196, opened 2023-04-28
(dsl): Support source filtering in search requests

mvelimir:


modules/library/src/main/scala/zio/elasticsearch/ElasticRequest.scala
```diff
@@ -552,45 +556,59 @@ object ElasticRequest {
       extends ElasticRequest[SearchResult]
       with HasFrom[SearchRequest]
       with HasRouting[SearchRequest]
+      with HasSize[SearchRequest]
       with HasSort[SearchRequest]
-      with HasSize[SearchRequest] {
+      with HasSourceFiltering[SearchRequest] {
     def aggregate(aggregation: ElasticAggregation): SearchAndAggregateRequest
 
+    def excludes(fields: String*): SearchRequest
```
drmarjanovic, line 564: Do we want to support `excludes(field: String, fields: String*)`?

modules/library/src/main/scala/zio/elasticsearch/ElasticRequest.scala
```diff
@@ -552,45 +556,59 @@ object ElasticRequest {
...
+
     def highlights(value: Highlights): SearchRequest
 
+    def includes(fields: String*): SearchRequest
```
drmarjanovic, line 568: Do we want to support `includes(field: String, fields: String*)`?

modules/library/src/main/scala/zio/elasticsearch/request/options/HasSourceFiltering.scala
```diff
@@ -0,0 +1,42 @@
...
+package zio.elasticsearch.request.options
+
+import zio.schema.Schema
+
+trait HasSourceFiltering[R <: HasSourceFiltering[R]] {
+  def excludes(fields: String*): R
+
+  def includes(fields: String*): R
+
+  final def includes(schema: Schema.Record[_]): R = {
```
drmarjanovic, line 26: Can we annotate this with `@tailrec`?

mvelimir: With this implementation it's not possible due to the `flatMap` call on the list, and I'm not sure it is possible with a different implementation either, because each element might call the recursive function.

modules/library/src/main/scala/zio/elasticsearch/ElasticRequest.scala
arnoldlacko: I think we'll have to choose one or the other as having both will probably cause `ambiguity` as `excludes(oneValue)` satisfies both. I think the decision should come down to whether we think the list will be used in a dynamic manner, in which case the stricter form can be less convenient.

modules/library/src/main/scala/zio/elasticsearch/ElasticRequest.scala
```diff
@@ -552,45 +556,59 @@ object ElasticRequest {
...
+
     def searchAfter(value: Json): SearchRequest
   }
 
   private[elasticsearch] final case class Search(
     index: IndexName,
     query: ElasticQuery[_],
     sortBy: Set[Sort],
+    excluded: Option[List[String]],
```
arnoldlacko, line 577: I think a `Chunk` would be more appropriate here, especially since we rely on concatenation in `def excludes`, even though most likely `excludes` will not be used multiple times.

modules/library/src/main/scala/zio/elasticsearch/ElasticRequest.scala
```diff
@@ -615,7 +633,16 @@ object ElasticRequest {
...
 
-      fromJson merge sizeJson merge highlightsJson merge sortJson merge self.query.toJson merge searchAfterJson
+      val sourceJson: Json =
+        included
+          .map(included => Obj("includes" -> Arr(included.map(_.toJson): _*)))
+          .map { included =>
+            included merge excluded.fold(Obj())(excluded => Obj("excludes" -> Arr(excluded.map(_.toJson): _*)))
+          }
+          .orElse(excluded.map(excluded => Obj("excludes" -> Arr(excluded.map(_.toJson): _*))))
+          .fold(Obj())(filters => Obj("_source" -> filters))
```
arnoldlacko, line 643: I wonder if something like this could be a bit easier to read:
```
val sourceJson: Json = 
   (included, excluded) match {
      case (None, None) => Obj()
      case (in, ex) => 
         Obj( "_source" -> {
            val includes = included.fold(Obj())(included => Obj("includes" -> Arr(included.map(_.toJson): _*)))
            val excludes = excluded.fold(Obj())(excluded => Obj("excludes" -> Arr(excluded.map(_.toJson): _*))
            included merge excluded
         }
      }
```
Would this work?

modules/library/src/main/scala/zio/elasticsearch/request/options/HasSourceFiltering.scala
arnoldlacko, line 26: Would it make sense (or be possible) to have something like
```
includeOnly[A](implicit schema: Schema.Record[A])
```
?

modules/library/src/main/scala/zio/elasticsearch/ElasticRequest.scala
```diff
@@ -671,13 +707,23 @@ object ElasticRequest {
       val sortJson: Json =
         if (self.sortBy.nonEmpty) Obj("sort" -> Arr(self.sortBy.toList.map(_.paramsToJson): _*)) else Obj()
 
+      val sourceJson: Json =
```
dbulaja98, line 710: Same as above. (Arnold's comment)

modules/library/src/main/scala/zio/elasticsearch/request/options/HasSourceFiltering.scala
```diff
@@ -0,0 +1,42 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package zio.elasticsearch.request.options
+
+import zio.schema.Schema
+
+trait HasSourceFiltering[R <: HasSourceFiltering[R]] {
```
dbulaja98, line 21: I think we can make this trait `private[elasticsearch]`?

dbulaja98: Also we should write `scaladoc` for methods in this trait.

modules/library/src/test/scala/zio/elasticsearch/ElasticRequestDSLSpec.scala
```diff
@@ -116,12 +197,18 @@ object ElasticRequestDSLSpec extends ZIOSpecDefault {
...
+          .excludes() match {
           case r: ElasticRequest.Search => r.toJson
         }
         val expected =
           """
             |{
+            |  "_source" : {
+            |    "includes" : [],
+            |    "excludes" : []
+            |  },
```
dbulaja98, line 211: Should we allow this?

drmarjanovic (approved):

PR #205, opened 2023-05-06
Change return type of executing bulk request

jlcanela:
A good practice with ElasticSearch when bulk insert is to verify the status for each Bulk response item so that it’s possible to retry this item associated request if necessary.  Current bulk request returning Unit prevents such practice.

This PR:
* updates executeBulk to return a BulkResponse instead of Unit
  * new BulkResponse case class in zio.elasticsearch.executor.response package
  * Note: all related case classes are in the same file, maybe you’d rather have each case class in its own file ?
* updates tests
  * HttpElasticExecutorSpec, now ensure a BulkResponse is returned instead of Unit
  * it/HttpExecutorSpec, now ensure BulkResponse returns bulk request status including error case


modules/library/src/main/scala/zio/elasticsearch/executor/HttpExecutor.scala
```diff
@@ -44,7 +45,7 @@ import zio.elasticsearch.request.{CreationOutcome, DeletionOutcome, UpdateOutcom
 import zio.elasticsearch.result._
 import zio.json.ast.Json
 import zio.json.ast.Json.{Arr, Obj, Str}
-import zio.json.{DeriveJsonDecoder, JsonDecoder}
+import zio.json.{DeriveJsonDecoder, JsonDecoder, _}
```
drmarjanovic, line 48: Does it make sense to use `import zio.json._`?

modules/library/src/main/scala/zio/elasticsearch/executor/response/BulkResponse.scala
```diff
@@ -0,0 +1,172 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package zio.elasticsearch.executor.response
+
+import zio.json.{DeriveJsonDecoder, JsonDecoder, jsonField, jsonHint}
+
+// Bulk response format:
```
drmarjanovic, line 21: I guess we can remove this comment. 

Also, can we keep only attributes that are meaningful for the library response? For example, I would remove `seqNo`, `primaryTerm`, etc.

drmarjanovic (commented):
Hey @jlcanela, 

I would like to thank you for your contribution! Amazing work overall! 

Besides several remarks in the code, I would like to point out that `BulkResponse` is a package private class, and therefore, I'm not sure it can be used from the app - you can try it in our example app.

We have `private[elasticsearch]` for internal Elastic responses, and our practice is to have a separate class for the library responses - you can check the `result` package. We should identify what is the most important info for the bulk response and define `BulkResult` which will be public.

modules/library/src/main/scala/zio/elasticsearch/executor/HttpExecutor.scala
jlcanela: executeBulk implementation is based on sendRequest implementation, it’s zio.json._ to parse body into BulkResponse. 
=> a better solution is to use sendRequestWithCustomerResponse

modules/library/src/main/scala/zio/elasticsearch/executor/response/BulkResponse.scala
jlcanela: ok I’ll do that

jlcanela:
> Hey @jlcanela,
> 
> I would like to thank you for your contribution! Amazing work overall!
> 
> Besides several remarks in the code, I would like to point out that `BulkResponse` is a package private class, and therefore, I'm not sure it can be used from the app - you can try it in our example app.
> 
> We have `private[elasticsearch]` for internal Elastic responses, and our practice is to have a separate class for the library responses - you can check the `result` package. We should identify what is the most important info for the bulk response and define `BulkResult` which will be public.

Thank you for your kind remarks. 
I now use similar patterns to 'result' package. Don’t hesitate if you have any comment.

modules/library/src/main/scala/zio/elasticsearch/executor/HttpExecutor.scala
```diff
@@ -126,18 +127,26 @@ private[elasticsearch] final class HttpExecutor private (esConfig: ElasticConfig
       }
     }
 
-  private def executeBulk(r: Bulk): Task[Unit] = {
+  private def executeBulk[A](r: Bulk): Task[BulkResponse] = {
```
arnoldlacko, line 130: Is there any usage of the type parameter `A`?

modules/library/src/main/scala/zio/elasticsearch/executor/response/ShardsResponse.scala
```diff
@@ -0,0 +1,29 @@
...
+
+package zio.elasticsearch.executor.response
+
+import zio.json.{DeriveJsonDecoder, JsonDecoder}
+
+final case class ShardsResponse private[elasticsearch] (
+  total: Int,
+  successful: Int,
+  failed: Int
+)
```
dbulaja98, line 25: Can you use already defined final case class `Shards`?  it has same fields as `ShardsResponse` + `skipped` field.

modules/library/src/main/scala/zio/elasticsearch/executor/response/Status.scala
```diff
@@ -0,0 +1,28 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package zio.elasticsearch.executor.response
+
+import zio.json.{DeriveJsonDecoder, JsonDecoder}
+
+final case class Status private[elasticsearch] (
```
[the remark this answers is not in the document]
dbulaja98: Also, if yes, can it be `private[elasticsearch]`?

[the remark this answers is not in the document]
dbulaja98: Also, if yes, can it be `private[elasticsearch]`?

modules/library/src/main/scala/zio/elasticsearch/executor/HttpExecutor.scala
jlcanela: removed in next commit.

modules/library/src/main/scala/zio/elasticsearch/executor/response/ShardsResponse.scala
jlcanela: updated next commit.

modules/library/src/main/scala/zio/elasticsearch/executor/response/Status.scala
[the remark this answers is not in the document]
jlcanela: Totally right, it’s some code I forgot to remove as status in Item is an Int.

drmarjanovic (approved):

PR #215, opened 2023-05-11
Replace List with Chunk

### LINKED ISSUE #190: Swap List[_] with Chunk[_]
Swap usage of List collection with Chunk in all places, except if there would be significant hit on performance.

mvelimir:
Resolves #190

### DIFF
--- a/modules/library/src/main/scala/zio/elasticsearch/result/ElasticResult.scala
+++ b/modules/library/src/main/scala/zio/elasticsearch/result/ElasticResult.scala
@@ -20,7 +20,7 @@ import zio.elasticsearch.executor.response.{AggregationResponse, SearchWithAggre
 import zio.json.ast.Json
 import zio.prelude.ZValidation
 import zio.schema.Schema
-import zio.{IO, Task, UIO, ZIO}
+import zio.{Chunk, IO, Task, UIO, ZIO}
 
 private[elasticsearch] sealed trait AggregationsResult {
   def aggregation(name: String): Task[Option[AggregationResponse]]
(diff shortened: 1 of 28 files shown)


dbulaja98 (approved):

modules/library/src/main/scala/zio/elasticsearch/result/ElasticResult.scala
```diff
@@ -58,17 +58,17 @@ final class GetResult private[elasticsearch] (private val doc: Option[Item]) ext
...
   def documentAs[A: Schema]: IO[DecodingException, List[A]] =
     ZIO.fromEither {
-      ZValidation.validateAll(hits.map(item => ZValidation.fromEither(item.documentAs))).toEitherWith { errors =>
+      ZValidation.validateAll(hits.map(item => ZValidation.fromEither(item.documentAs)).toList).toEitherWith { errors =>
         DecodingException(s"Could not parse all documents successfully: ${errors.map(_.message).mkString(", ")}")
       }
     }
 
-  lazy val items: UIO[List[Item]] = ZIO.succeed(hits)
+  lazy val items: UIO[List[Item]] = ZIO.succeed(hits.toList)
```
markaya, line 71: Why we don't return UIO[Chunk[Item]]?

modules/library/src/main/scala/zio/elasticsearch/result/ElasticResult.scala
```diff
@@ -91,14 +91,14 @@ final class SearchAndAggregateResult private[elasticsearch] (
 
   def documentAs[A: Schema]: Task[List[A]] =
```
markaya, line 92: Same here, Task[Chunk[A]]

modules/library/src/main/scala/zio/elasticsearch/result/Item.scala
```diff
@@ -46,14 +46,15 @@ final case class Item(
   def innerHitAs[A](name: String)(implicit schema: Schema[A]): Either[DecodingException, List[A]] =
```
markaya, line 46: Can we return Either[DecodingException, Chunk[A]] here?

markaya (changes requested):

markaya (approved):

drmarjanovic (approved):

PR #216, opened 2023-05-11
Create public methods for script and inner hits

markaya:
Summary: 

Implemented idea for creating `ElasticUtilities` object that would contain methods to create objects such as `Script` or `InnerHits` that do not belong to any other objects.


drmarjanovic (changes requested):
Why not use a companion object? If it does not work right now, we can make only the constructor private and create it through the `apply` method.

dbulaja98 (approved):

modules/library/src/main/scala/zio/elasticsearch/query/InnerHits.scala
```diff
@@ -41,12 +41,14 @@ private[elasticsearch] final case class InnerHits(
...
 object InnerHits {
+  def empty: InnerHits = InnerHits(from = None, name = None, size = None)
+
   def from(value: Int): InnerHits =
-    InnerHits(from = Some(value))
+    InnerHits(from = Some(value), name = None, size = None)
 
-  def name(value: String): InnerHits =
-    InnerHits(name = Some(value))
+  def withName(value: String): InnerHits =
```
drmarjanovic, line 49: Why is this changed? Not sorted.

modules/library/src/main/scala/zio/elasticsearch/script/Script.scala
```diff
@@ -57,6 +57,6 @@ private[elasticsearch] final case class Script(
 }
 
 object Script {
-  def apply(source: String): Script =
+  def from(source: String): Script =
```
drmarjanovic, line 60: Why use `from` instead of `apply`?

drmarjanovic (changes requested):

modules/library/src/main/scala/zio/elasticsearch/script/Script.scala
markaya: We agreed earlier that overriding `apply` method is not something we want to be doing as it makes false feeling that Script has only one field.

modules/library/src/main/scala/zio/elasticsearch/query/InnerHits.scala
drmarjanovic: I think we are trying to avoid `with` prefix. @arnoldlacko?

modules/library/src/main/scala/zio/elasticsearch/script/Script.scala
drmarjanovic: I would consider `source` as a required param. And if we have `Script("source")` then we can consider it as a wrapper on the string. I would use `apply`.

modules/library/src/main/scala/zio/elasticsearch/query/InnerHits.scala
markaya: Maybe we could drop all these and just have `InnerHits.empty.name(...).from(...)`?

modules/library/src/main/scala/zio/elasticsearch/script/Script.scala
markaya: Sure. I will change it to `apply`.

drmarjanovic (approved):

PR #219, opened 2023-05-12
Remove `paramsToJson` methods

dbulaja98:


modules/library/src/main/scala/zio/elasticsearch/ElasticRequest.scala
```diff
@@ -628,16 +634,15 @@ object ElasticRequest {
...
 
     def toJson: Json = {
-      val fromJson: Json = self.from.fold(Obj())(f => Obj("from" -> f.toJson))
+      val fromJson: Json = from.fold(Obj())(f => Obj("from" -> f.toJson))
 
-      val sizeJson: Json = self.size.fold(Obj())(s => Obj("size" -> s.toJson))
+      val sizeJson: Json = size.fold(Obj())(s => Obj("size" -> s.toJson))
 
-      val highlightsJson: Json = highlights.map(_.toJson).getOrElse(Obj())
+      val highlightsJson: Json = highlights.map(h => Obj("highlight" -> h.toJson)).getOrElse(Obj())
```
drmarjanovic, line 641: Let's use `.fold` instead of `.getOrElse(Obj())`.

modules/library/src/main/scala/zio/elasticsearch/ElasticRequest.scala
```diff
@@ -709,16 +720,15 @@ object ElasticRequest {
...
```
drmarjanovic, line 727: Same here.

drmarjanovic (changes requested):

modules/library/src/main/scala/zio/elasticsearch/ElasticRequest.scala
```diff
@@ -645,7 +650,13 @@ object ElasticRequest {
             })
         }
 
-      fromJson merge sizeJson merge highlightsJson merge sortJson merge self.query.toJson merge searchAfterJson merge sourceJson
+      fromJson merge
+        sizeJson merge
+        highlightsJson merge
+        sortJson merge
+        Obj("query" -> query.toJson(fieldPath = None)) merge
```
markaya, line 657: If it makes no difference, stylistically I would put `Obj("query" -> query.toJson(fieldPath = None))` on first place.

markaya (changes requested):

modules/library/src/main/scala/zio/elasticsearch/executor/HttpExecutor.scala
```diff
@@ -156,8 +156,9 @@ private[elasticsearch] final class HttpExecutor private (esConfig: ElasticConfig
       .get(uri"${esConfig.uri}/${r.index}/$Count".withParams(getQueryParams(Chunk(("routing", r.routing)))))
       .contentType(ApplicationJson)
       .response(asJson[CountResponse])
+      .body()
```
drmarjanovic, line 159: Why do you need this?

dbulaja98: It should not be there.

drmarjanovic (approved):

PR #231, opened 2023-05-26
Make Aggregations API public

markaya:
Summary: 
Made classes used in aggregations public so users can use them to access aggregations and sub aggeragatoins.


dbulaja98 (commented):
I would rather leave this `private[elasticsearch]` and create another "layer" - AggregationResult which we will return

PR #257, opened 2023-06-24
Add support for min aggregation

Mitep:


modules/library/src/main/scala/zio/elasticsearch/package.scala
```diff
@@ -67,6 +60,18 @@ package object elasticsearch extends IndexNameNewtype with RoutingNewtype {
...
       zio.flatMap(_.aggregations)
 
+    /**
+     * Executes the [[ElasticRequest.SearchRequest]] or the [[ElasticRequest.SearchAndAggregateRequest]].
+     *
+     * @param name
+     *   the name of the aggregation to retrieve
+     * @return
+     *   a [[RIO]] effect that, when executed, will produce the aggregation as instance of
+     *   [[result.AvgAggregationResult]].
```
dbulaja98, line 70: Check if this compiles without warnings.

modules/library/src/test/scala/zio/elasticsearch/ElasticAggregationSpec.scala
```diff
@@ -141,6 +141,19 @@ object ElasticAggregationSpec extends ZIOSpecDefault {
             equalTo(Max(name = "aggregation", field = "intField", missing = Some(20.0)))
           )
         },
+        test("min") {
```
dbulaja98, line 144: Can you add aggregationWithAllParams? For encoding too.

dbulaja98 (changes requested):

modules/library/src/test/scala/zio/elasticsearch/ElasticAggregationSpec.scala
Mitep: Test already includes case classes where we cover all fields.

modules/library/src/main/scala/zio/elasticsearch/package.scala
Mitep: I don't see any warning regarding RIO, only a couple regarding unused methods.

dbulaja98 (approved):

docs/overview/aggregations/elastic_aggregation_min.md
```diff
@@ -0,0 +1,35 @@
...
+val aggregation: MinAggregation = minAggregation(name = "minAggregation", field = "intField")
+```
+
+You can create a [type-safe](https://lambdaworks.github.io/zio-elasticsearch/overview/overview_zio_prelude_schema) `Min` aggregation using the `minAggregation` method this way:
+```scala
+// Document.intField must be number value, because of Min aggregation
+val aggregation: MinAggregation = minAggregation(name = "minAggregation", field = Document.intField)
+```
+
+If you want to change the `missing`, you can use `missing` method:
```
markaya, line 25:
```suggestion
If you want to change the `missing` parameter, you can use `missing` method:
```

markaya (changes requested):

markaya (approved):

PR #282, opened 2023-07-06
Allow ElasticPrimitive types in `term` and `terms` query

drmarjanovic:


modules/library/src/main/scala/zio/elasticsearch/ElasticQuery.scala
```diff
@@ -699,12 +699,14 @@ object ElasticQuery {
...
    *   text value that will be used for the query
+   * @tparam A
+   *   the type of value to be matched. A JSON decoder must be in scope for this type
    * @tparam S
    *   document for which field query is executed
    * @return
    *   an instance of [[zio.elasticsearch.query.TermQuery]] that represents the term query to be performed.
    */
-  final def term[S](field: Field[S, String], value: String): TermQuery[S] =
+  final def term[S, A: ElasticPrimitive](field: Field[S, A], value: A): TermQuery[S] =
```
dbulaja98, line 709: What do you think about leaving just `_` instead of giving strict `A: ElasticPrimitive`?

drmarjanovic: We use the same approach for `matches` query.

drmarjanovic: I mean, we are limiting possible types for values. I wouldn't say we should allow passing an object as the argument.

mvelimir (approved):

Issue #91, opened 2023-02-08
Implement Query API


drmarjanovic:
### Queries

- [ ] `fuzzyQuery`
- [ ] `geoBoundingBoxQuery`
- [ ] `geoDistanceQuery`
- [ ] `geoPolygonQuery`
- [ ] `geoShapeQuery`
- [ ] `hasChildQuery`
- [ ] `hasParentQuery`
- [ ] `idsQuery`
- [ ] `intervalsQuery`
- [ ] `matchBooleanPrefixQuery`
- [ ] `matchPhraseQuery`
- [ ] `matchPhrasePrefixQuery`
- [ ] `multiMatchQuery`
- [ ] `nestedQuery`
- [ ] `prefixQuery`
- [ ] `regexpQuery`
- [ ] `scriptQuery`
- [ ] `scriptScoreQuery`
- [ ] `shapeQuery`
- [ ] `simpleQueryStringQuery`
- [ ] `termsQuery`
- [ ] `termsSetQuery`
- [ ] `queryStringQuery`

### Tips

- Find comprehensive info about the query you are going to implement [here](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/query-dsl.html)
- Define query in `zio.elasticsearch.ElasticQuery` companion object
- Define the public method that represents DSL for the query in the `zio.elasticsearch.ElasticQuery` companion object
- Support additional parameters such as "boost", "case insensitive", etc. in the `zio.elasticsearch` package
- Provide unit tests in the `zio.elasticsearch.QueryDSLSpec` to confirm behavior
- Provide integration tests (`it` module) in the `zio.elasticsearch.HttpExectorSpec` to confirm the behavior
- 😇 Feel free to extend other tests as well
- 😊 Feel free to address any follow-up issue either for the query you are implementing or implemented one already

_**Note:** You can use #48 as a reference. If you feel that the pull request size is growing out of control, feel free to split it but make sure to link this issue in each of the related PRs._

petarcurcin:
I'm taking prefixQuery

LeonaNedeljkovic:
I'm taking geoPolygonQuery

bblagojevic94:
I'm taking `geoShapeQuery`

filipmilo:
Taking `idsQuery`

PR #305, opened 2023-08-17
(dsl): Support `multiMatch` query

milicns:
Part of #91
 
Fields parameter is made optional so implementation of this query is a bit different than others.

And should `type` parameter be included and made optional? https://www.elastic.co/guide/en/elasticsearch/reference/7.17/query-dsl-multi-match-query.html#multi-match-types


modules/library/src/main/scala/zio/elasticsearch/ElasticQuery.scala
```diff
@@ -447,6 +447,18 @@ object ElasticQuery {
...
+   * [[zio.elasticsearch.query.MultiMatchQuery]] query builds on the `match` query to allow multi-field queries.
+   *
+   * @param value
+   *   the text value to be matched
+   * @return
+   *   an instance of [[zio.elasticsearch.query.MultiMatchQuery]] that represents the multi match query to be performed.
+   */
+  final def multiMatch(value: String): MultiMatchQuery[Any] =
+    MultiMatch(fields = None, value = value)
+
```
milicns, line 461: Because fields are optional and type-safe method is now in `HasFields` trait, i suppose we don't need `multiMatch` type-safe method?

modules/library/src/main/scala/zio/elasticsearch/query/Queries.scala
```diff
@@ -584,6 +585,25 @@ private[elasticsearch] final case class MatchPhrase[S](field: String, value: Str
     )
 }
 
+sealed trait MultiMatchQuery[S] extends ElasticQuery[S] with HasFields[MultiMatchQuery[S]]
+
+private[elasticsearch] final case class MultiMatch[S](fields: Option[Chunk[String]], value: String)
```
dbulaja98, line 590: I would also add `minimum_should_match` and `boost` parameters, as they are already implemented, and can be applied to this query.

docs/overview/queries/elastic_query_multi_match.md
```diff
@@ -0,0 +1,30 @@
+---
+id: elastic_query_multi_match
+title: "Multi Match Query"
+---
+
+The `MultiMatch` query builds on the `match` query to allow multi-field queries:
```
dbulaja98, line 6: This seems to be unfinished?

modules/library/src/main/scala/zio/elasticsearch/ElasticQuery.scala
dbulaja98: If you change `fields` inside of query, this will then be: `fields=Chunk.empty`

modules/library/src/main/scala/zio/elasticsearch/query/Queries.scala
```diff
@@ -584,6 +585,25 @@ private[elasticsearch] final case class MatchPhrase[S](field: String, value: Str
...
+    extends MultiMatchQuery[S] { self =>
+  def fields(fields: String*): MultiMatchQuery[S] =
+    self.copy(fields = Some(Chunk.fromIterable(fields)))
+
+  def fields[S1: Schema](fields: Field[S1, String]*): MultiMatchQuery[S] =
+    self.copy(fields = Some(Chunk.fromIterable(fields.map(_.toString))))
```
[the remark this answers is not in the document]
dbulaja98: Also, some sort of restrictions should be discussed, as (I think) we can not have both type-safe and raw fields. @arnoldlacko what do you think?

dbulaja98 (changes requested):

modules/library/src/main/scala/zio/elasticsearch/query/Queries.scala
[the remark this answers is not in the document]
arnoldlacko: I think given that we use `_.toString` here, the resulting query is the same regardless of being constructed using Field or raw strings, so it does not need to be an issue if they are mixed.
What I wonder is why we have a `S1` type here with no defined relation to `S`, and why it does not affect the type of the resulting query.
@mvelimir any thoughts?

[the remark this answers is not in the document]
mvelimir: We'll definitely need `S1` to be a subtype of `S`, to narrow the type which will always start from `Any` based on the constructor method. Also, I think mixing both type-safe fields and plain strings should be allowed, since strings in this case can be used as wildcards, and having type-safe fields will still be useful in inferring the type of the document. Also, the type-safe method should definitely contain at least one field as to not clash with the strings method when the argument list is empty. With all this in mind, I think the method signature should be:
```scala
def fields[S1 <: S: Schema](field: Field[S1, String], fields: Field[S1, String]*): MultiMatchQuery[S1]
```

modules/library/src/main/scala/zio/elasticsearch/query/Queries.scala
arnoldlacko: I think it could make sense to allow something like `.matchingType(MultiMatchType.Phrase)`.
If it's more complex, it could also be a separate PR.

drmarjanovic (changes requested):

modules/library/src/main/scala/zio/elasticsearch/query/Queries.scala
```diff
@@ -26,6 +27,8 @@ import zio.schema.Schema
 
 import scala.annotation.unused
 
+import MultiMatchType.MultiMatchType
```
dbulaja98, line 30: Can you omit this import?

modules/library/src/main/scala/zio/elasticsearch/query/Queries.scala
```diff
@@ -593,6 +596,82 @@ private[elasticsearch] final case class MatchPhrasePrefix[S](field: String, valu
...
+   *   a new instance of the [[zio.elasticsearch.query.ElasticQuery]] with the `fields` array set.
+   */
+  def fields(field: String, fields: String*): MultiMatchQuery[S]
+
+  /**
+   * Sets the type-safe `fields` parameter for this [[zio.elasticsearch.query.ElasticQuery]]. The `fields` parameter is
+   * the type-safe array of fields that will be searched.
+   *
+   * @param fields
+   *   a type-safe array of fields to set `fields` parameter to
```
dbulaja98, line 620:
```suggestion
   *   a array of type-safe fields to set `fields` parameter to
```

modules/library/src/main/scala/zio/elasticsearch/query/Queries.scala
```diff
@@ -593,6 +596,82 @@ private[elasticsearch] final case class MatchPhrasePrefix[S](field: String, valu
...
+   * @param fields
+   *   a array of fields to set `fields` parameter to
+   * @return
+   *   a new instance of the [[zio.elasticsearch.query.ElasticQuery]] with the `fields` array set.
+   */
+  def fields(field: String, fields: String*): MultiMatchQuery[S]
+
+  /**
+   * Sets the type-safe `fields` parameter for this [[zio.elasticsearch.query.ElasticQuery]]. The `fields` parameter is
+   * the type-safe array of fields that will be searched.
```
dbulaja98, line 617:
```suggestion
   * the array of type-safe fields that will be searched.
```

modules/library/src/main/scala/zio/elasticsearch/query/Queries.scala
```diff
@@ -593,6 +596,82 @@ private[elasticsearch] final case class MatchPhrasePrefix[S](field: String, valu
...
+    with HasMinimumShouldMatch[MultiMatchQuery[S]] {
+
+  /**
+   * Sets the `fields` parameter for this [[zio.elasticsearch.query.ElasticQuery]]. The `fields` parameter is the array
+   * of fields that will be searched.
+   *
+   * @param fields
+   *   a array of fields to set `fields` parameter to
+   * @return
+   *   a new instance of the [[zio.elasticsearch.query.ElasticQuery]] with the `fields` array set.
```
dbulaja98, line 611:
```suggestion
   *   an instance of [[zio.elasticsearch.query.ElasticQuery]] enriched with the `fields` parameter.
```

modules/library/src/main/scala/zio/elasticsearch/query/Queries.scala
```diff
@@ -593,6 +596,82 @@ private[elasticsearch] final case class MatchPhrasePrefix[S](field: String, valu
...
+  def fields(field: String, fields: String*): MultiMatchQuery[S]
+
+  /**
+   * Sets the type-safe `fields` parameter for this [[zio.elasticsearch.query.ElasticQuery]]. The `fields` parameter is
+   * the type-safe array of fields that will be searched.
+   *
+   * @param fields
+   *   a type-safe array of fields to set `fields` parameter to
+   * @return
+   *   a new instance of the [[zio.elasticsearch.query.ElasticQuery]] with the `fields` array set.
```
dbulaja98, line 622: Same as above.

modules/library/src/main/scala/zio/elasticsearch/query/Queries.scala
```diff
@@ -593,6 +596,82 @@ private[elasticsearch] final case class MatchPhrasePrefix[S](field: String, valu
...
+   */
+  def fields[S1 <: S: Schema](field: Field[S1, String], fields: Field[S1, String]*): MultiMatchQuery[S1]
+
+  /**
+   * Sets the `type` parameter for this [[zio.elasticsearch.query.ElasticQuery]]. The `type` parameter decides the way
+   * [[zio.elasticsearch.query.ElasticQuery]] is executed internally.
+   *
+   * @param matchingType
+   *   the [[zio.elasticsearch.query.MultiMatchType]] value of 'type' parameter
```
dbulaja98, line 631: Can you explain possible values for `matchingType`.
You can lookup for other places with similar @param.

modules/library/src/main/scala/zio/elasticsearch/query/Queries.scala
```diff
@@ -593,6 +596,82 @@ private[elasticsearch] final case class MatchPhrasePrefix[S](field: String, valu
...
+   * @return
+   *   a new instance of the [[zio.elasticsearch.query.ElasticQuery]] with the `type` parameter set.
```
dbulaja98, line 633: Same as above (an instance..).

modules/library/src/main/scala/zio/elasticsearch/query/Queries.scala
```diff
@@ -593,6 +596,82 @@ private[elasticsearch] final case class MatchPhrasePrefix[S](field: String, valu
...
+  boost: Option[Double],
+  matchingType: Option[MultiMatchType],
+  minimumShouldMatch: Option[Int]
+) extends MultiMatchQuery[S] { self =>
+
+  def boost(boost: Double): MultiMatchQuery[S] =
+    self.copy(boost = Some(boost))
+
+  def fields(field: String, fields: String*): MultiMatchQuery[S] =
+    self.copy(fields = Chunk.fromIterable(field +: fields))
```
dbulaja98, line 650: Does it make sense to add fields to chunk instead of setting it? @drmarjanovic

modules/library/src/test/scala/zio/elasticsearch/ElasticQuerySpec.scala
```diff
@@ -897,6 +897,109 @@ object ElasticQuerySpec extends ZIOSpecDefault {
...
+          val query                       = multiMatch("this is a test")
+          val queryWithFields             = multiMatch("this is a test").fields("stringField1", "stringField2")
+          val queryWithFieldsTs           = multiMatch("this is a test").fields(TestDocument.stringField)
+          val queryWithFieldsSuffix       = multiMatch("this is a test").fields(TestDocument.stringField.raw)
+          val queryWithType               = multiMatch("this is a test").matchingType(MultiMatchType.BestFields)
+          val queryWithBoost              = multiMatch("this is a test").boost(2.2)
+          val queryWithMinimumShouldMatch = multiMatch("this is a test").minimumShouldMatch(2)
+          val queryWithAllParams = multiMatch("this is a test")
+            .fields(TestDocument.stringField)
+            .matchingType(MultiMatchType.BestFields)
```
dbulaja98, line 910: Can you omit `MultiMatchType`? (everywhere in file)

modules/library/src/test/scala/zio/elasticsearch/ElasticQuerySpec.scala
```diff
@@ -2505,13 +2608,120 @@ object ElasticQuerySpec extends ZIOSpecDefault {
               |    "stringField": {
               |       "query" : "test"
               |    }
-              |  }
+              |   }
```
dbulaja98, line 2611: Was this on purpose?

dbulaja98 (changes requested):

modules/library/src/main/scala/zio/elasticsearch/query/MultiMatchType.scala
```diff
@@ -0,0 +1,27 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package zio.elasticsearch.query
+
+private[elasticsearch] object MultiMatchType extends Enumeration {
```
dbulaja98, line 19: Please refer to the existing classes that represents enumerations.

dbulaja98 (changes requested):

modules/library/src/main/scala/zio/elasticsearch/query/Queries.scala
```diff
@@ -608,18 +606,18 @@ sealed trait MultiMatchQuery[S]
...
    * @return
-   *   a new instance of the [[zio.elasticsearch.query.ElasticQuery]] with the `fields` array set.
+   *   an instance of the [[zio.elasticsearch.query.ElasticQuery]] enriched with the `fields` parameter.
    */
   def fields(field: String, fields: String*): MultiMatchQuery[S]
 
   /**
    * Sets the type-safe `fields` parameter for this [[zio.elasticsearch.query.ElasticQuery]]. The `fields` parameter is
-   * the type-safe array of fields that will be searched.
+   * the array of type-safe fields that will be searched.
```
dbulaja98, line 615: Omit `the`.

modules/library/src/main/scala/zio/elasticsearch/query/Queries.scala
```diff
@@ -628,9 +626,15 @@ sealed trait MultiMatchQuery[S]
...
    *
    * @param matchingType
-   *   the [[zio.elasticsearch.query.MultiMatchType]] value of 'type' parameter
+   *   the [[zio.elasticsearch.query.MultiMatchType]] value of 'type' parameter, possible values are:
+   *   - [[zio.elasticsearch.query.MultiMatchType.BestFields]]
+   *   - [[zio.elasticsearch.query.MultiMatchType.BoolPrefix]]
+   *   - [[zio.elasticsearch.query.MultiMatchType.CrossFields]]
+   *   - [[zio.elasticsearch.query.MultiMatchType.MostFields]]
+   *   - [[zio.elasticsearch.query.MultiMatchType.Phrase]]
+   *   - [[zio.elasticsearch.query.MultiMatchType.PhrasePrefix]]
```
dbulaja98, line 635: Please, provide short explanation for each value. You can see `scaladoc` for `scoreMode` def.

modules/library/src/main/scala/zio/elasticsearch/query/Queries.scala
```diff
@@ -660,10 +664,11 @@ private[elasticsearch] final case class MultiMatch[S](
...
     val multiMatchFields =
-      matchingType.map("type" -> _.toJson) ++ (if (fields.nonEmpty) Some("fields" -> Arr(fields.map(_.toJson)))
-                                               else None) ++ boost.map("boost" -> _.toJson) ++ minimumShouldMatch.map(
-        "minimum_should_match" -> _.toJson
-      )
+      matchingType.map("type" -> _.value.toJson) ++ (if (fields.nonEmpty) Some("fields" -> Arr(fields.map(_.toJson)))
+                                                     else None) ++ boost.map("boost" -> _.toJson) ++ minimumShouldMatch
+        .map(
+          "minimum_should_match" -> _.toJson
+        )
```
dbulaja98, line 671: Can this be in one line? Or better formatted?

dbulaja98 (changes requested):

modules/library/src/main/scala/zio/elasticsearch/query/MultiMatchType.scala
```diff
@@ -16,12 +16,16 @@
...
+}
+
+object MultiMatchType {
+  case object BestFields   extends MultiMatchType { def value: String = "best_fields"   }
+  case object BoolPrefix   extends MultiMatchType { def value: String = "bool_prefix"   }
+  case object CrossFields  extends MultiMatchType { def value: String = "cross_fields"  }
+  case object MostFields   extends MultiMatchType { def value: String = "most_fields"   }
+  case object Phrase       extends MultiMatchType { def value: String = "phrase_fields" }
+  case object PhrasePrefix extends MultiMatchType { def value: String = "phrase_prefix" }
 }
```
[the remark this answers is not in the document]
milicns: If you mean at the end of file, sbt prepare removes it

modules/library/src/main/scala/zio/elasticsearch/query/MultiMatchType.scala
```diff
@@ -16,16 +16,14 @@
...
+sealed trait MultiMatchType
 
 object MultiMatchType {
-  case object BestFields   extends MultiMatchType { def value: String = "best_fields"   }
-  case object BoolPrefix   extends MultiMatchType { def value: String = "bool_prefix"   }
-  case object CrossFields  extends MultiMatchType { def value: String = "cross_fields"  }
-  case object MostFields   extends MultiMatchType { def value: String = "most_fields"   }
-  case object Phrase       extends MultiMatchType { def value: String = "phrase_fields" }
-  case object PhrasePrefix extends MultiMatchType { def value: String = "phrase_prefix" }
+  case object BestFields   extends MultiMatchType { override def toString: String = "best_fields"   }
```
dbulaja98, line 22: Can you please put defs to new line inside of objects?

dbulaja98 (changes requested):

dbulaja98 (approved):

PR #368, opened 2023-11-17
Refactor AggregationResponse

vanjaftn:


modules/library/src/main/scala/zio/elasticsearch/executor/response/AggregationResponse.scala
```diff
@@ -116,51 +116,12 @@ private[elasticsearch] object AvgAggregationResponse {
...
-private[elasticsearch] final case class FilterAggregationResponse(
-  @jsonField("doc_count")
-  docCount: Int,
-  subAggregations: Option[Map[String, AggregationResponse]] = None
-) extends AggregationResponse
-
-private[elasticsearch] object FilterAggregationResponse extends JsonDecoderOps {
-  implicit val decoder: JsonDecoder[FilterAggregationResponse] = Obj.decoder.mapOrFail { case Obj(fields) =>
-    val allFields = fields.flatMap { case (field, data) =>
+private[elasticsearch] sealed trait BucketDecoder extends JsonDecoderOps {
```
dbulaja98, line 119: Can you please add 'JsonDecoderOps' in this sealed trait too?

modules/library/src/main/scala/zio/elasticsearch/executor/response/AggregationResponse.scala
```diff
@@ -350,107 +358,13 @@ private[elasticsearch] final case class TermsAggregationBucket(
...
   implicit val decoder: JsonDecoder[TermsAggregationBucket] = Obj.decoder.mapOrFail { case Obj(fields) =>
-    val allFields = fields.flatMap { case (field, data) =>
-      field match {
-        case "key" =>
-          Some(field -> data.toString.replaceAll("\"", ""))
-        case "doc_count" =>
-          Some(field -> data.unsafeAs[Int])
-        case _ =>
-          val objFields = data.unsafeAs[Obj].fields.toMap
+    val allFields = BucketDecoder(fields).allFields
```
dbulaja98, line 363: Because we are using `BucketDecoder` this way, maybe we don't even need this class to be `implicit` and then there is no need for `sealed trait`.

dbulaja98: If we do this, we should probably revert changes about `JsonDecoderOps`.

dbulaja98 (changes requested):

dbulaja98 (commented):
Also rebase this PR (introduce changes from percentile ranks aggregation).

modules/library/src/main/scala/zio/elasticsearch/executor/response/AggregationResponse.scala
```diff
@@ -118,6 +118,112 @@ private[elasticsearch] object AvgAggregationResponse {
   implicit val decoder: JsonDecoder[AvgAggregationResponse] = DeriveJsonDecoder.gen[AvgAggregationResponse]
 }
 
+case class BucketDecoder(fields: Chunk[(String, Json)]) extends JsonDecoderOps {
```
dbulaja98, line 121: Make this case class private.

modules/library/src/main/scala/zio/elasticsearch/executor/response/AggregationResponse.scala
```diff
@@ -159,110 +265,12 @@ private[elasticsearch] final case class FilterAggregationResponse(
...
-            case str if str.contains("terms#") =>
-              Some(field -> data.unsafeAs[TermsAggregationResponse](TermsAggregationResponse.decoder))
-            case str if str.contains("value_count#") =>
-              Some(field -> ValueCountAggregationResponse(value = objFields("value").unsafeAs[Int]))
-          }
-      }
-    }.toMap
+    val allFields = BucketDecoder(fields).allFields
+    val docCount  = allFields("doc_count").asInstanceOf[Int]
+    val subAggs   = BucketDecoder(fields).subAggs
```
dbulaja98, line 272: You are creating this two times `BucketDecoder(fields)`. Extract this in val before these three vals.

modules/library/src/main/scala/zio/elasticsearch/executor/response/AggregationResponse.scala
```diff
@@ -366,113 +374,13 @@ private[elasticsearch] final case class TermsAggregationBucket(
...
-            case str if str.contains("sum#") =>
-              Some(field -> SumAggregationResponse(value = objFields("value").unsafeAs[Double]))
-            case str if str.contains("terms#") =>
-              Some(field -> data.unsafeAs[TermsAggregationResponse](TermsAggregationResponse.decoder))
-            case str if str.contains("value_count#") =>
-              Some(field -> ValueCountAggregationResponse(value = objFields("value").unsafeAs[Int]))
-          }
-      }
-    }.toMap
+    val allFields = BucketDecoder(fields).allFields
```
dbulaja98, line 379: Here also.

dbulaja98 (changes requested):

dbulaja98 (approved):

PR #391, opened 2024-02-15
(example): Update zio-config to use standard approach

drmarjanovic:


modules/example/src/main/scala/example/Main.scala
```diff
@@ -30,7 +31,10 @@ import scala.util.Using
 
 object Main extends ZIOAppDefault {
 
-  override def run: Task[ExitCode] = {
+  override val bootstrap: Layer[Nothing, Unit] =
```
arnoldlacko, line 34: Could be shortened to `ULayer[Unit]` :)

arnoldlacko (approved):

dbulaja98 (dismissed):

arnoldlacko (approved):

PR #554, opened 2025-01-25
Bump zio-prelude to 1.0.0-RC37

drmarjanovic:
Closes #543


mvelimir (changes requested):
You should also change the Scala badge in the README to 2.13.16

PR #648, opened 2025-06-13
Added QueryStringQuery

JuliaIlic:


drmarjanovic (changes requested):
Also, additionally:
- Please use better PR title
- Link the issue you're resolving in description
- Format your code before pushing it - you can use `sbt prepare`

dbulaja98 (changes requested):
Good first PR.

Just few things: 
- You are missing some integrations tests
- You are missing documentation page for this query

modules/library/src/main/scala/zio/elasticsearch/ElasticQuery.scala
```diff
@@ -899,6 +899,48 @@ object ElasticQuery {
...
+   * @param fields
+   *   the type-safe fields to be searched
+   * @param query
+   *   the query string to search for
+   * @tparam S
+   *   the document type on which the query is executed
+   * @return
+   *   an instance of [[zio.elasticsearch.query.QueryStringQuery]] that represents the query to be performed.
+   */
+  final def queryStringQuery[S: Schema](query: String, fields: Field[S, _]*): QueryStringQuery[S] =
```
drmarjanovic, line 935: Does it make sense for fields to be empty?

modules/library/src/main/scala/zio/elasticsearch/query/Queries.scala
```diff
@@ -1016,6 +1016,49 @@ private[elasticsearch] final case class Prefix[S](
...
+  minimumShouldMatch: Option[Int]
+) extends QueryStringQuery[S] { self =>
+
+  def boost(value: Double): QueryStringQuery[S] =
+    self.copy(boost = Some(value))
+
+  def fields(field: String, fields: String*): QueryStringQuery[S] =
+    self.copy(fields = Chunk.fromIterable(field +: fields))
+
+  def fields(fields: Chunk[String]): QueryStringQuery[S] =
```
drmarjanovic, line 1038: Do we need this?

modules/library/src/main/scala/zio/elasticsearch/query/Queries.scala
```diff
@@ -1016,6 +1016,49 @@ private[elasticsearch] final case class Prefix[S](
...
+    self.copy(fields = fields)
+
+  def minimumShouldMatch(value: Int): QueryStringQuery[S] =
+    self.copy(minimumShouldMatch = Some(value))
+
+  private[elasticsearch] def toJson(fieldPath: Option[String]): Json = {
+    val fieldsJson =
+      if (fields.nonEmpty) Some("fields" -> Arr(fields.map(Str(_))))
```
drmarjanovic, line 1046: Can we use `.map` here?

drmarjanovic (changes requested):
Several things are missing:
- website docs
- integration tests

Besides that, let's rename a method and link to the issue properly.

PR #840, opened 2026-09-18
(dsl): Support `queryString` query

drmarjanovic:
Resolves #648


modules/library/src/main/scala/zio/elasticsearch/query/Queries.scala
```diff
@@ -1016,6 +1016,53 @@ private[elasticsearch] final case class Prefix[S](
...
 }
 
+sealed trait QueryStringQuery[S]
+    extends ElasticQuery[S]
+    with HasFields[QueryStringQuery, S]
+    with HasBoost[QueryStringQuery[S]]
+    with HasMinimumShouldMatch[QueryStringQuery[S]]
+
+private[elasticsearch] final case class QueryString[S](
+  defaultField: Option[String],
```
dbulaja98, line 1026: `defaultField` is modeled and serialized, but there's no public way to set it: both `queryString` builders in `ElasticQuery.scala` hardcode `defaultField` = `None` and `QueryString` itself is `private[elasticsearch]`, so users can't set it via copy either. Should we add a `defaultField(value: String)` builder method to `QueryStringQuery`, like the other params have?

docs/overview/queries/elastic_query_query_string.md
```diff
@@ -0,0 +1,60 @@
...
+If you want to define the `minimum_should_match` parameter, use the `minimumShouldMatch` method:
+```scala
+val query: QueryStringQuery[Any] =
+queryString(query = "name").minimumShouldMatch(2)
+```
+
+You can also construct the query manually with all parameters:
+```scala
+val query: QueryStringQuery[Document] =
+QueryString(
```
dbulaja98, line 53: `QueryString` is `private[elasticsearch]`, so this snippet won't actually compile for library users. Same issue as the existing `elastic_query_simple_query_string.md` doc. Maybe show construction via the `queryString(...)` builder instead.

modules/library/src/main/scala/zio/elasticsearch/query/Queries.scala
```diff
@@ -1016,6 +1016,53 @@ private[elasticsearch] final case class Prefix[S](
...
+
+  def minimumShouldMatch(value: Int): QueryStringQuery[S] =
+    self.copy(minimumShouldMatch = Some(value))
+
+  private[elasticsearch] def toJson(fieldPath: Option[String]): Json = {
+    val fieldsJson =
+      if (fields.nonEmpty) Some("fields" -> Arr(fields.map(Str(_))))
+      else None
+
+    val params = Chunk(
```
dbulaja98, line 1053: This builds JSON with raw `Str(_)` / `Json.Num(_)` instead of the `.toJson` extension methods other queries in this file use (e.g. Prefix, Term). Worth aligning for consistency, though not a functional bug.

dbulaja98 (changes requested):

modules/library/src/main/scala/zio/elasticsearch/query/Queries.scala
```diff
@@ -1016,6 +1016,68 @@ private[elasticsearch] final case class Prefix[S](
...
+   * @param value
+   *   the [[scala.Predef.String]] value for `default_field` parameter
+   * @return
+   *   an instance of the [[zio.elasticsearch.query.QueryStringQuery]] enriched with the `default_field` parameter.
+   */
+  def defaultField(value: String): QueryStringQuery[S]
+}
+
+private[elasticsearch] final case class QueryString[S](
+  defaultField: Option[String],
```
dbulaja98: `defaultField` and `fields` can now both be set on `QueryStringQuery`, but `Elasticsearch`'s `query_string` query explicitly rejects that combination (`QueryStringQueryBuilder` throws `cannot use [fields] parameter in conjunction with [default_field]`). Worse, the PR's own new docs example (`elastic_query_query_string.md`, "combine all parameters") demonstrates exactly this broken combination: `queryString(...).fields(...).defaultField(...)`. This should either be documented as invalid, or the type-level API should prevent it.

modules/library/src/main/scala/zio/elasticsearch/ElasticQuery.scala
```diff
@@ -899,6 +899,50 @@ object ElasticQuery {
   final def prefix(field: String, value: String): Prefix[Any] =
     Prefix(field = field, value = value, caseInsensitive = None)
 
+  /**
+   * Constructs an instance of [[zio.elasticsearch.query.QueryStringQuery]] using the specified parameters.
```
dbulaja98: `queryString` still says it "supports query strings with simple syntax," copy-pasted from `simpleQueryString`; it should describe the strict Lucene-like syntax, as the markdown docs correctly do.

modules/library/src/test/scala/zio/elasticsearch/ElasticQuerySpec.scala
```diff
@@ -4000,6 +4075,82 @@ object ElasticQuerySpec extends ZIOSpecDefault {
           assert(query.toJson(fieldPath = None))(equalTo(expected.toJson)) &&
           assert(queryWithCaseInsensitive.toJson(fieldPath = None))(equalTo(expectedWithCaseInsensitive.toJson))
         },
+        test("queryString") {
```
dbulaja98: The JSON-encoding test for `queryString` never asserts `default_field` serialization, unlike the constructing test above it.

modules/library/src/main/scala/zio/elasticsearch/query/Queries.scala
drmarjanovic: Fixed in 7a46645: `defaultField` and `fields` are now mutually exclusive. Setting `defaultField` clears `fields`, and every `fields` overload clears `defaultField`, so the query can never serialize both. This is documented in the `defaultField` Scaladoc and the markdown docs, and the "combine all parameters" example no longer mixes the two. Tests cover both orderings.

---
_Generated by [Claude Code](https://claude.ai/code)_

modules/library/src/main/scala/zio/elasticsearch/ElasticQuery.scala
drmarjanovic: Fixed in 7a46645. Both `queryString` overloads now describe the strict Lucene-like syntax: operators, wildcards, fuzzy matching and field names in the query string, and an error on invalid syntax.

---
_Generated by [Claude Code](https://claude.ai/code)_

modules/library/src/test/scala/zio/elasticsearch/ElasticQuerySpec.scala
drmarjanovic: Added in 7a46645. The JSON test now asserts `default_field` serialization on its own and together with `boost`/`minimum_should_match`.

---
_Generated by [Claude Code](https://claude.ai/code)_

docs/overview/queries/elastic_query_query_string.md
drmarjanovic: Addressed in 7f9e9b1. The docs now build every example through the public `queryString(...)` builders instead of the private `QueryString` case class.

---
_Generated by [Claude Code](https://claude.ai/code)_

modules/library/src/main/scala/zio/elasticsearch/query/Queries.scala
drmarjanovic: Addressed in 7f9e9b1. `QueryString.toJson` now uses the `.toJson` extension methods (`query.toJson`, `_.toJson` for fields, boost and `minimum_should_match`), like `Prefix` and `Term`.

---
_Generated by [Claude Code](https://claude.ai/code)_

dbulaja98 (approved):

PR #841, opened 2026-09-18
(dsl): Support `intervalQuery` query

drmarjanovic:
Resolves #647


modules/library/src/main/scala/zio/elasticsearch/query/ElasticIntervalRule.scala
```diff
@@ -0,0 +1,472 @@
...
+
+  def lt(value: String): IntervalRangeRule[S] = self.copy(upper = Some(LessThanInterval(value)))
+
+  def lte(value: String): IntervalRangeRule[S] = self.copy(upper = Some(LessThanOrEqualToInterval(value)))
+
+  def useField(field: Field[_, _]): IntervalRangeRule[S] = self.copy(useField = Some(field.name))
+
+  def useField(field: String): IntervalRangeRule[S] = self.copy(useField = Some(field))
+
+  private[elasticsearch] def toJson: Json =
```
dbulaja98, line 387: `IntervalRange.toJson` always emits "gte"/"lte" regardless of whether the bound was built with `.gt()`/`.lt()` (exclusive) or `.gte()`/`.lte()` (inclusive). Since none of the `GreaterThanInterval`/`LessThanInterval` cases carry their own key, `intervalRange.gt("10").lt("20")` serializes to `{"gte":"10","lte":"20"}` instead of `{"gt":"10","lt":"20"}`, silently including boundary values the caller meant to exclude. The existing `Range`/`LowerBound`/`UpperBound` pattern in `Queries.scala` emits the correct key per case, this should follow the same approach.

modules/library/src/main/scala/zio/elasticsearch/query/ElasticIntervalRule.scala
```diff
@@ -0,0 +1,472 @@
...
+  useField: Option[String]
+) extends IntervalRegexpRule[S] { self =>
+
+  def analyzer(value: String): IntervalRegexpRule[S] = self.copy(analyzer = Some(value))
+
+  def useField(field: Field[_, _]): IntervalRegexpRule[S] = self.copy(useField = Some(field.name))
+
+  def useField(field: String): IntervalRegexpRule[S] = self.copy(useField = Some(field))
+
+  private[elasticsearch] def toJson: Json =
```
dbulaja98, line 433: `IntervalRegexp.toJson` calls `pattern.toJson(None)` on the full `RegexpQuery`, which renders the entire query wrapper `({"regexp":{"<field>":{"value":...}}})` and nests that under "pattern". This produces `{"regexp":{"pattern":{"regexp":{...}}}}` instead of the plain regex string Elasticsearch expects `({"regexp":{"pattern":"a.*",...}})`. This sends a malformed intervals query for any use of `intervalRegexp`, and there's no test asserting its JSON output.

modules/library/src/main/scala/zio/elasticsearch/query/ElasticIntervalRule.scala
```diff
@@ -0,0 +1,472 @@
...
+   *   a new instance of the interval rule with the `transpositions` parameter set to `true`.
+   */
+  def transpositionsEnabled: IntervalFuzzyRule[S]
+}
+
+private[elasticsearch] final case class IntervalFuzzy[S](
+  term: String,
+  prefixLength: Option[Int],
+  transpositions: Option[Boolean],
+  fuzziness: Option[String],
```
dbulaja98, line 187: `fuzziness` is defined on `IntervalFuzzy` and serialized in `toJson`, but `IntervalFuzzyRule` never exposes a builder method to set it. Since `IntervalFuzzy` is `private[elasticsearch]`, callers have no way to construct or copy it directly either, `fuzziness` is permanently stuck at `None`. Should add a `fuzziness(...)` method on `IntervalFuzzyRule`.

dbulaja98 (changes requested):

modules/library/src/main/scala/zio/elasticsearch/query/ElasticIntervalRule.scala
```diff
@@ -0,0 +1,491 @@
...
+
+  def fuzziness(value: String): IntervalFuzzyRule[S] = self.copy(fuzziness = Some(value))
+
+  def prefixLength(length: Int): IntervalFuzzyRule[S] = self.copy(prefixLength = Some(length))
+
+  def transpositionsDisabled: IntervalFuzzyRule[S] = self.copy(transpositions = Some(false))
+
+  def transpositionsEnabled: IntervalFuzzyRule[S] = self.copy(transpositions = Some(true))
+
+  def useField(field: Field[_, _]): IntervalFuzzyRule[S] = self.copy(useField = Some(field.name))
```
dbulaja98: The typed `useField(field: Field[_, _])` overload serializes `field.name` (leaf) instead of `field.toString` (full dotted path). Breaks for nested fields like `address.city`, sends the wrong field name to ES. Untested.
5 more call sites: 284, 316, 395, 448, 477

modules/library/src/main/scala/zio/elasticsearch/query/options/HasUseField.scala
```diff
@@ -0,0 +1,42 @@
...
+
+  /**
+   * Sets the `use_field` parameter for this [[zio.elasticsearch.query.IntervalRule]].
+   *
+   * @param field
+   *   the type-safe field to use from the document definition
+   * @return
+   *   a new instance of the query with the `use_field` value set.
+   */
+  def useField(field: Field[_, _]): Q
```
dbulaja98: `useField`'s typed overload isn't constrained to the rule's document type `S`, so it silently accepts a field from an unrelated document type. Existing `HasFields[Q[_], S]` pattern elsewhere does this correctly.

modules/library/src/main/scala/zio/elasticsearch/query/ElasticIntervalRule.scala
drmarjanovic: Fixed in bb39bcd: each `IntervalRangeBound` case now carries its own `key` (`gt`/`gte`/`lt`/`lte`), and `IntervalRange.toJson` emits it. Covered by the `intervalRange with exclusive bounds` test.

---
_Generated by [Claude Code](https://claude.ai/code)_

modules/library/src/main/scala/zio/elasticsearch/query/ElasticIntervalRule.scala
drmarjanovic: Fixed in bb39bcd: `IntervalRegexp` now holds the pattern as a plain `String` and serializes `{"regexp":{"pattern":"..."}}`. The `intervalRegexp` test asserts the JSON output.

---
_Generated by [Claude Code](https://claude.ai/code)_

modules/library/src/main/scala/zio/elasticsearch/query/ElasticIntervalRule.scala
drmarjanovic: Fixed in bb39bcd: added `fuzziness(value: String)` to `IntervalFuzzyRule`. The `intervalFuzzy` test covers it.

---
_Generated by [Claude Code](https://claude.ai/code)_

modules/library/src/main/scala/zio/elasticsearch/query/ElasticIntervalRule.scala
drmarjanovic: Fixed in 16c3ad6: all six typed `useField` overloads now serialize `field.toString`, so the full dotted path is kept. The new `intervalRule with type-safe useField` test covers every rule, including a nested field (`nestedField.stringField`).

---
_Generated by [Claude Code](https://claude.ai/code)_

modules/library/src/main/scala/zio/elasticsearch/query/options/HasUseField.scala
drmarjanovic: Fixed in 16c3ad6: `HasUseField` is now `HasUseField[Q[_], S]`, following `HasFields`. The typed overload is `useField[S1 <: S: Schema](field: Field[S1, _]): Q[S1]`, so a field from an unrelated document no longer compiles. The `Schema` bound blocks the `S1 = Nothing` escape that `Field`'s contravariance would otherwise allow. A `typeCheck` test covers both the accepted case and the rejected one.

---
_Generated by [Claude Code](https://claude.ai/code)_

modules/library/src/main/scala/zio/elasticsearch/ElasticQuery.scala
```diff
@@ -548,6 +548,37 @@ object ElasticQuery {
...
+   * @param field
+   *   the type-safe field on which the query is executed
+   * @param rule
+   *   an instance of [[zio.elasticsearch.query.IntervalRule]] representing the interval query rule
+   * @tparam S
+   *   the document type for which the query is defined
+   * @return
+   *   an [[zio.elasticsearch.ElasticQuery]] instance representing the intervals query.
+   */
+  final def intervals[S](field: Field[S, _], rule: IntervalRule): IntervalsQuery[S] = Intervals(field.toString, rule)
```
dbulaja98: `intervals[S](field: Field[S, _], rule: IntervalRule)` doesn't actually constrain rule to `S` - `IntervalRule` itself carries no type parameter, so the type-safety you just added to `useField` doesn't propagate to the query-construction call site. You can mix a field from one document type with a rule built/narrowed for a completely different one and it still compiles, only breaking at runtime against `Elasticsearch`.

drmarjanovic: Fixed in adc85b7: `IntervalRule` now carries the document type, following the `ElasticQuery[-S]` pattern.

- `IntervalRule[-S]` and all rule traits are contravariant in `S`.
- The untyped constructors (`intervalMatch`, `intervalPrefix`, …) return `[Any]`. The type-safe `useField` and `filter` narrow a rule to a specific document type. `intervalAllOf`, `intervalAnyOf` and `intervalFilter` carry through the type of the rules they combine.
- `intervals[S: Schema](field: Field[S, _], rule: IntervalRule[S])`: a rule narrowed to a different document type, whether directly or nested in `all_of`/`any_of`/`filter`, no longer compiles. The type is inferred as `A with B`, and there is no `Schema` for that, the same way `must`/`filter` work. The string-field variant accepts only `IntervalRule[Any]`.

New `typeCheck` tests cover the rejected cases (direct, inside `any_of`, inside `filter`) and a valid mix of typed and untyped rules.

---
_Generated by [Claude Code](https://claude.ai/code)_

modules/library/src/main/scala/zio/elasticsearch/query/ElasticIntervalRule.scala
```diff
@@ -0,0 +1,504 @@
...
+private[elasticsearch] final case class IntervalRange[S](
+  lower: Option[IntervalRangeBound],
+  upper: Option[IntervalRangeBound],
+  analyzer: Option[String],
+  useField: Option[String]
+) extends IntervalRangeRule[S] { self =>
+
+  def analyzer(value: String): IntervalRangeRule[S] = self.copy(analyzer = Some(value))
+
+  def gt(value: String): IntervalRangeRule[S] = self.copy(lower = Some(GreaterThanInterval(value)))
```
dbulaja98: `IntervalRange`'s `gt`/`gte` both write the same `lower` field, and `lt`/`lte` both write the same `upper` field, so calling both silently drops the first with no compile or runtime error.

modules/library/src/main/scala/zio/elasticsearch/query/ElasticIntervalRule.scala
```diff
@@ -0,0 +1,504 @@
...
+final case class IntervalFilter[-S](
+  after: Option[IntervalRule[S]] = None,
+  before: Option[IntervalRule[S]] = None,
+  containedBy: Option[IntervalRule[S]] = None,
+  containing: Option[IntervalRule[S]] = None,
+  notContainedBy: Option[IntervalRule[S]] = None,
+  notContaining: Option[IntervalRule[S]] = None,
+  notOverlapping: Option[IntervalRule[S]] = None,
+  overlapping: Option[IntervalRule[S]] = None,
+  script: Option[Json] = None
```
dbulaja98: `IntervalFilter.script` is typed as raw `Option[Json]` instead of the library's existing type-safe `zio.elasticsearch.script.Script` type used everywhere else scripting appears.

modules/library/src/main/scala/zio/elasticsearch/query/ElasticIntervalRule.scala
```diff
@@ -0,0 +1,504 @@
...
+          lower.map(bound => bound.key -> bound.toJson),
+          upper.map(bound => bound.key -> bound.toJson),
+          analyzer.map("analyzer" -> Str(_)),
+          useField.map("use_field" -> _.toJson)
+        ).flatten: _*
+      )
+    )
+}
+
+private[elasticsearch] sealed trait IntervalRangeBound {
```
dbulaja98: `IntervalRangeBound` is a sealed trait with four near-identical one-line case classes that each just fix a `key` string - could be one case class `IntervalRangeBound(key: String, value: String)`.

modules/library/src/main/scala/zio/elasticsearch/query/ElasticIntervalRule.scala
```diff
@@ -0,0 +1,504 @@
...
+    self.copy[S1](useField = Some(field.toString))
+
+  def useField(field: String): IntervalPrefixRule[S] = self.copy(useField = Some(field))
+
+  private[elasticsearch] def toJson: Json =
+    Obj(
+      "prefix" -> Obj(
+        Chunk(
+          Some("prefix" -> Str(prefix)),
+          analyzer.map(a => "analyzer" -> Str(a)),
```
dbulaja98: The `analyzer` field is serialized three different ways across the six rule case classes (`.map(a => "analyzer" -> Str(a))` vs `.map("analyzer" -> Str(_))` vs `.map("analyzer" -> _.toJson)`).

modules/library/src/main/scala/zio/elasticsearch/query/ElasticIntervalRule.scala
```diff
@@ -0,0 +1,504 @@
...
+}
+
+private[elasticsearch] final case class IntervalAllOf[S](
+  intervals: Chunk[IntervalRule[S]],
+  maxGaps: Option[Int],
+  ordered: Option[Boolean],
+  filter: Option[IntervalFilter[S]]
+) extends IntervalAllOfRule[S] { self =>
+
+  def filter[S1 <: S](f: IntervalFilter[S1]): IntervalAllOfRule[S1] = self.copy[S1](filter = Some(f))
```
dbulaja98: `filter[S1 <: S](f: IntervalFilter[S1])` is hand-copied identically three times (IntervalAllOf, IntervalAnyOf, IntervalMatch) with no shared trait, unlike `analyzer`/`useField` which were factored into `HasAnalyzer`/`HasUseField`.

modules/library/src/main/scala/zio/elasticsearch/query/Queries.scala
```diff
@@ -782,6 +780,21 @@ private[elasticsearch] final case class Ids[S](values: Chunk[String]) extends Id
...
 }
 
+sealed trait IntervalsQuery[S] extends ElasticQuery[S]
+
+private[elasticsearch] final case class Intervals[S](
+  field: String,
+  rule: IntervalRule[S]
+) extends IntervalsQuery[S] { self =>
+
+  private[elasticsearch] def toJson(fieldPath: Option[String]): Json =
```
dbulaja98: `Intervals.toJson(fieldPath)` prefixes the outer `field` with the enclosing nested-query path, but `IntervalRule.toJson` takes no fieldPath parameter, so a rule's `use_field` is never prefixed the same way.

modules/library/src/main/scala/zio/elasticsearch/ElasticIntervalRule.scala
```diff
@@ -0,0 +1,235 @@
...
+   * Constructs an instance of [[zio.elasticsearch.query.IntervalRule]] that matches terms containing the specified
+   * `pattern`.
+   *
+   * @param pattern
+   *   the substring that a matching term must contain
+   * @return
+   *   an instance of [[zio.elasticsearch.query.IntervalRule]] representing the `wildcard` interval rule.
+   */
+  def intervalContains(pattern: String): IntervalWildcardRule[Any] =
+    IntervalWildcard(s"*$pattern*", analyzer = None, useField = None)
```
dbulaja98: `intervalContains`/`intervalEndsWith`/`intervalStartsWith` build wildcard patterns via raw string interpolation without escaping Elasticsearch wildcard metacharacters `*`/`?` in the caller's pattern.

modules/library/src/main/scala/zio/elasticsearch/ElasticQuery.scala
```diff
@@ -548,6 +548,40 @@ object ElasticQuery {
...
+   *   the type-safe field on which the query is executed
+   * @param rule
+   *   an instance of [[zio.elasticsearch.query.IntervalRule]] representing the interval query rule, defined for the
+   *   same document type as the `field`
+   * @tparam S
+   *   the document type for which the query is defined. An implicit `Schema` instance must be in scope
+   * @return
+   *   an [[zio.elasticsearch.ElasticQuery]] instance representing the intervals query.
+   */
+  final def intervals[S: Schema](field: Field[S, _], rule: IntervalRule[S]): IntervalsQuery[S] =
```
dbulaja98: `intervals[S: Schema](field: Field[S,_], rule: IntervalRule[S])` requires an implicit `Schema[S]` that's never used in the body, unlike every sibling single-Field typed builder in the same file (`contains[S]`, `matchPhrase[S]`, `prefix[S]`, `wildcard[S]`), none of which declare a Schema bound.

modules/library/src/main/scala/zio/elasticsearch/query/ElasticIntervalRule.scala
drmarjanovic: Fixed in 936ef22. `IntervalRangeRule` now tracks its bounds in the type, the same way `RangeQuery` does (`IntervalRangeRule[-S, LB <: LowerBound, UB <: UpperBound]`), and `gt`/`gte` require `LB =:= Unbounded.type`, and `lt`/`lte` require `UB =:= Unbounded.type`. So `intervalRange.gt("1").gte("2")` no longer compiles. `typeCheck` tests cover this.

---
_Generated by [Claude Code](https://claude.ai/code)_

modules/library/src/main/scala/zio/elasticsearch/query/ElasticIntervalRule.scala
drmarjanovic: Fixed in 936ef22: `IntervalFilter.script` (and the `intervalFilter` parameter) is now `Option[zio.elasticsearch.script.Script]` and serializes via `Script.toJson`. A new test covers it.

---
_Generated by [Claude Code](https://claude.ai/code)_

modules/library/src/main/scala/zio/elasticsearch/query/ElasticIntervalRule.scala
drmarjanovic: Fixed in 936ef22, but by removing `IntervalRangeBound` entirely instead of collapsing it: the range rule now reuses the existing `GreaterThan`/`GreaterThanOrEqualTo`/`LessThan`/`LessThanOrEqualTo`/`Unbounded` bounds from `RangeQuery`. The typed bounds are also what the compile-time single-bound check relies on.

---
_Generated by [Claude Code](https://claude.ai/code)_

modules/library/src/main/scala/zio/elasticsearch/query/ElasticIntervalRule.scala
drmarjanovic: Fixed in 936ef22: `analyzer` and the other string values (`query`, `prefix`) are now all serialized with `_.toJson`.

---
_Generated by [Claude Code](https://claude.ai/code)_

modules/library/src/main/scala/zio/elasticsearch/query/ElasticIntervalRule.scala
drmarjanovic: Fixed in 936ef22: `filter` is now declared once in a new `HasIntervalFilter[Q[-_], -S]` trait, which `IntervalAllOfRule`, `IntervalAnyOfRule` and `IntervalMatchRule` extend, the same way `HasAnalyzer`/`HasUseField` work.

---
_Generated by [Claude Code](https://claude.ai/code)_

modules/library/src/main/scala/zio/elasticsearch/query/Queries.scala
drmarjanovic: Fixed in 936ef22: `IntervalRule.toJson` (and `IntervalFilter.toJson`) now takes `fieldPath`. `Intervals` passes it down through `all_of`/`any_of`/`filter`, and `use_field` gets the same prefix as the intervals field. A new test covers `intervals` inside `nested`, including a `use_field` on a rule nested in a filter.

---
_Generated by [Claude Code](https://claude.ai/code)_

modules/library/src/main/scala/zio/elasticsearch/ElasticIntervalRule.scala
drmarjanovic: Fixed in 936ef22: `intervalContains`/`intervalStartsWith`/`intervalEndsWith` now backslash-escape `\`, `*` and `?` in the caller's value, so it's matched literally. The scaladoc says so, and a test covers it. `intervalWildcard` still takes a raw pattern. The top-level `ElasticQuery.contains`/`startsWith`/`endsWith` don't escape either, but I left them unchanged to keep this PR scoped. Happy to follow up separately if you'd like them aligned.

---
_Generated by [Claude Code](https://claude.ai/code)_

modules/library/src/main/scala/zio/elasticsearch/ElasticQuery.scala
drmarjanovic: Keeping this one, because the `Schema` bound is what enforces the type safety from the previous comment. Unlike `contains`/`matchPhrase`/`prefix`/`wildcard`, `intervals` combines two typed inputs, a field and a rule. Both `Field[-S, _]` and `IntervalRule[-S]` are contravariant. Without the bound, `intervals(TestDocument.stringField, rule: IntervalRule[TestSubDocument])` compiles with `S` inferred as `TestDocument with TestSubDocument`. The implicit `Schema[S]` is what rejects that intersection. This is the same mechanism `must`/`filter`/`should` and `HasFields.fields` use when combining typed inputs. Removing it makes the `intervals accepts only rules defined for the field's document type` tests fail. In 936ef22 I expanded the scaladoc to say why the bound is there.

---
_Generated by [Claude Code](https://claude.ai/code)_

dbulaja98 (approved):

